PR #33, opened 2016-09-05
Add missing LICENSE from Far 2

### LINKED ISSUE #19: Licensing clarity
I noticed that there is no `LICENSE` file that is present in standard Far3 repository. This is a license violation that can lead to termination of `far2l` project o GitHub.

To avoid "license deadlock" and make it more useful for upstream, the original BSD code pieces could be kept separated from GPLv2 both in headers and in top level LICENSE file.

techtonik:
Closes #19 after consulting with https://opensource.stackexchange.com/questions/4424/remove-bsd-license-file-while-importing-code-into-gpl-project


elfmz:
Original far's BSD license notice still in the head each of FAR's source code that were  licensed under such license.
No sence in putting BSD license in the root of projects.  They're not BSD, they're GPL.

elfmz:
However FYI - colorer, netbox that also in the tree - they're natively GPL
WinPort also GPL due to its roots. So putting such license into the root of code is completely wrong. Putting into far2l - is ok, if noted that only some files initially BSD. Cuz for instance vtansi code derived from project that has zlib license, that is ok to used under GPL, but not OK to use under BSD.

techtonik:
I think that the point of conflict here is in the LICENSE text. Technically it is the same text as in headers, but LICENSE file is a separate copyright notice related to the whole project, so point 1 here:
```
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
```
is speaking specifically about LICENSE file itself, and about about source file headers.

techtonik:
> GPL the only one umbrella-license that compatible with all of used.

"Essentially, when it comes to the mingled form of the work, all the licenses applies, and the most restrictive one "wins". Meaning: in a compiled program with a mixture of BSD and GPL modules, the whole thing falls under the GPL." (c)  http://opensource.stackexchange.com/a/4429/6053

But that doesn't meant that old license can be superceded by a new one according:

"In a combination of programs under lax licenses, each part carries the license it came with. When the code is merged to the point that the parts can't be distinguished any more, that merged code should carry all the licenses of the merged parts." (c) Richard Stallman @ https://www.gnu.org/licenses/license-compatibility.html

People keep adding new answers - you should check this - https://opensource.stackexchange.com/questions/4424/remove-bsd-license-file-while-importing-code-into-gpl-project

elfmz:
There is no words about whole project, only source code. So select form any of:
1) far2l/LICENSE.Far2.txt prependent with alist of files to which it applies/
2) Split far2l to original code and all new added files (vt*.cpp/h). And put far2l/derived/LICENSE.Far2.txt
3) Leave as is.

techtonik:
This answer - http://opensource.stackexchange.com/a/4425/6053 - contains some strategies how to handle licenses so that users won't get misleaded. Quoting here:
```
* I could create a top level LICENSE or COPYING file that would explain at a high level the various licenses in use, and keep separate files for each licenses, either in a top level dir or a doc or licenses dir, and point to these in my top level file.
* I could merge it all in one big file (which could become somewhat hard to read and grok)
* I could use a more structured approach like in Debian copyright files, or my own https://github.com/nexB/attributecode tool using a structured yaml or yaml-like description of the various licenses used.
```

elfmz:
whell, approach with COPYING is OK. I will put it with links to GPL (at root) and to far2l/LICENSE.Far2.txt (placed in the folders where  its) with clarification in beginning which files it affects.

elfmz:
or you can do it, so this will be faster)

techtonik:
Open source is not only GPL. It is a lot of people who are invented licenses to get some attribution for their works. We can endlessly debate about how to avoid copyrights in a project, but at some point I'd like to see far2l packaged in Debian -> Ubuntu with changes pulled back to upstream projects. It is a cool initiative and it would be sad to see fail due to some awkward copyright issues, like it already happened with Wal Commander.

Let me move `LICENSE.Far2.txt` to `far2` then. =)

techtonik:
Done. =)

elfmz:
WAL commander had completely different problem - AFAIK originally author didn't specify license in the very beginning (I guessed this since [older version of its site](http://web.archive.org/web/20150428063559/http://www.wal-commander.org) doesnt contain License page), since that moment another guy forked project under MIT. Then original author put restrictive license on its sources, that prevents any forking. So in such case  there were mistakes of both authors. May be I'm wrong and there was restrictive license all the time in the source tree (unfortunatelly web.archive.org doesn't save that). But I don't think someone forked under MIT something that is clearly restricted against any forks..

elfmz:
OK if I will merge this but later will prepend list of affected files in the beginning? Or better split source tree of far2l/far2l on originals/added? Don't like 2nd option - it will introduce unneccessary mess...

techtonik:
Splitting source tree is an overkill, and maybe even a file list is an overkill. But if projects are able to maintain file list, it is of course awesome - https://github.com/spyder-ide/spyder/blob/master/LICENSE - it is even possible to maintain it automatically by parsing headers.

PR #161, opened 2016-10-21
Multiarc destination arc folder update

Gravityzwell:
Multiarc has "%%R - Current archive folder" variable to specify destination archive sub-folder.  Unfortunately, it is always quoted, so combining it to form a path fails:
```
%%R/%%f
"fold name"/file name
```

This pull adds a new %%r variable (lowercase). It adds the current archive folder unquoted, and if the value is not empty, it appends a "/" - notice the quotes are added in the manually:
```
"%%r%%f"
"fold name/file name"
```
This allows adding to zpaq and probably other archives in sub folders.

### DIFF
--- a/multiarc/src/arccmd.cpp
+++ b/multiarc/src/arccmd.cpp
@@ -344,6 +344,12 @@ int ArcCommand::ReplaceVar(char *Command,int &Length)
         }
       }
       break;
+    case 'r':
+      *Command=0;
+      if (RealArcDir && *RealArcDir) {
+        FSF.sprintf(Command,"%s/",RealArcDir);
+      }
+      break;
     case 'R':
       strcpy(Command,RealArcDir);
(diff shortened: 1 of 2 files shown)


elfmz:
May be better to remove quotes from %%R? Anyway they can be added manually..

Gravityzwell:
Removing quotes alone from %%R isn't enough, although I think quotes should indeed be added manually in the INI.

In the case of zpaq an additional trailing slash is needed, but only if %%R is not null.
```
"%%R/%%f"
"fold name/file name"  << okay %%R defined

"%%R/%%f"
"/file name"  << not okay, zpaq and maybe others treat this different than "file name"

"%%r/%%f"
"file name"  << okay %%r was null, slash omitted
```
I have not tested if just changing %%R to the new %%r behavior would allow RAR and CAB to keep working.

PR #194, opened 2016-11-07
Simplify README.md to use `make`

igorklopov:
To avoid using IDE. Probably most users will just need to make binaries without opening the project in IDE.


elfmz:
why you removed cmake pkg-config? Its still required for making, and not related to IDE.

igorklopov:
i added them, not removed. they were absent imho

PR #310, opened 2017-12-27
Introduced editorcomp

MikeMirzayanov:
Implemented simple Editor Autocompletion Plugin (editorcomp). It analyzes all tokens in a current file and suggests an ending for currently typing word. Use TAB to confirm a suggestion, just continue to type if you don't want to use a suggestion. The plugin can be completely disabled via "Plugin configuration".

You can easily change it to be disabled by default changing Editors::DEFAULT_ENABLED. But I prefer it is enabled by default. For me it is vital feature to write configuration/source files/natural language texts quickly and without mistypes.


SerjCh:
have tested currently on ubuntu 16.04.3 with issues:
1) completion with "tab" doesn't work
2) after first suggestion - next suggestions aren't showing anymore
3) in editor menu item Editor Autocomplete do nothing

atm have same plug on win, with tab completion and bs (or just typing/moving cursor) removal suggestion
win version of plugin have some options, screen attached
image

SerjCh:
from plugins configuration menu
image

SerjCh:
btw post with windows builds, may be anywhere there are sources of this builds
https://forum.farmanager.com/viewtopic.php?p=132507#p132507

MikeMirzayanov:
Added a commit but continuous integration seems to be broken.

SerjCh, what do you mean by "completion with tab doesn't work"? How did you test other items if it doesn't work?

BTW, I've installed ubuntu 16.04.3 64 bit and it works to me.

SerjCh:
out

MikeMirzayanov, I just try to use tab and nothing happens - just tab inserted (two spaces), next suggestion means that after first - no suggestions anymore, u can see this in gif animation. I had this bug on windows, fixed after downloaded johndows build on forum.

PR #314, opened 2018-01-27
typo fix

thinkingerrol:


thinkingerrol:
would be nice if there was some kind of chat for contributors, I would like to become one. Here just fixed a typo, what I would like to do is make sure far2l is easy to install on MacOS as pre-built binary and also from source, to be able to debug it on MacOS. If there is no existing chat, please join one I created just now for this purpose: https://t.me/joinchat/AAAAAFATvFp5wpDy-QdxsQ

PR #327, opened 2018-03-13
Add custom console RGB palette support

Artiomio:
Здравствуйте!
The 32 RGB values are stored in ~/.config/far2l/palette.ini and may be
changed. Palette config is initialized with defaults if palette.ini is
absent in the directory. An error message is shown in the console if
an error occurs while parsing the .ini file.
Parsing is done with https://github.com/benhoyt/inih.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,13 @@
-CMakeLists.txt.user
-
+CMakeCache.txt
+CMakeFiles
+CMakeScripts
+Testing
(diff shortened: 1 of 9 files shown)


elfmz:
need to change to existing ini file reader implemented in utils/KeyFileHelper.h

Artiomio:
Thanks for correcting! Works fine.

Issue #359, opened 2018-08-28
Support /etc/paths and /etc/paths.d/ on Mac OS


singalen:
Currently, if I run far2l as a stand-alone app from /Applications (created with the code in #332), its `$PATH` is a bare system-default `/usr/bin:/bin:/usr/sbin:/sbin`, even though `/etc/paths` and `/etc/paths.d/` are set up perfectly.

Thus, child processes - like archivers installed by MacPorts into `/opt/local/bin` - cannot be found, even though `/opt/local/bin` is in `/etc/paths`.

[There is no method works to set an app's default path](https://apple.stackexchange.com/questions/106355/setting-the-system-wide-path-environment-variable-in-mavericks), as of now. Looks like far2l has to set its `PATH` by itself.

elfmz:
fixed

PR #363, opened 2018-09-14
#359: Support /etc/paths and /etc/paths.d/ on Mac OS

singalen:
The code quality is kind of irrecoverable, so I mostly copy-pasted the `POpen()` function.

I'm not proud of it, but it does the job and is definitely better than inlining the logic.


elfmz:
will definately need to rewrite without using cat/tr - just to simple file read

PR #380, opened 2018-10-18
Simplify macOS install process

yurikoles:


elfmz:
IMHO its better to leave both ways. Say, easy and make all by you own hands.

yurikoles:
@elfmz good point! I had updated my commit, you can check-out result [here](https://github.com/yurikoles/far2l/blob/brew/README.md).

elfmz:
also since you changed build dir from outer location then add to .gitignore:
`build/`
..to avoid git status garbaging

yurikoles:
done

PR #385, opened 2018-10-19
Check against nullptr - nothrow added

svost:
https://en.cppreference.com/w/cpp/memory/new/nothrow

### DIFF
--- a/far2l/dialog.cpp
+++ b/far2l/dialog.cpp
@@ -3104,8 +3104,7 @@ int Dialog::ProcessKey(int Key)
 					return TRUE;
 				}
 
-				if (!(Item[FocusPos]->Flags & DIF_READONLY) ||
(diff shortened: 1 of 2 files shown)


far2l/dialog.cpp
```diff
@@ -3104,8 +3104,7 @@ int Dialog::ProcessKey(int Key)
 					return TRUE;
 				}
 
-				if (!(Item[FocusPos]->Flags & DIF_READONLY) ||
-				        ((Item[FocusPos]->Flags & DIF_READONLY) && IsNavKey(Key)))
+				if (!(Item[FocusPos]->Flags & DIF_READONLY) || IsNavKey(Key))
```
elfmz: неа, замена не эквивалентна

elfmz: а хотя...)

PR #427, opened 2018-11-14
cosmetic

svost:


multiarc/src/libpcre/study.c
```diff
@@ -323,7 +323,7 @@ do
           for (c = 0; c < 16; c++) start_bits[c] |= tcode[c];
           for (c = 128; c < 256; c++)
             {
-            if ((tcode[c/8] && (1 << (c&7))) != 0)
+            if ((tcode[c/8] & (1 << (c&7))) != 0)
```
elfmz: i totally agree that this looks as more correct, but on libpcre source code there is still &&: https://github.com/vmg/libpcre/blob/master/pcre_study.c#L1332

PR #442, opened 2018-12-07
some warnings on Apple

singalen:
There are more where these came from, but these are the most prominent.

### DIFF
--- a/far2l/wrap.cpp
+++ b/far2l/wrap.cpp
@@ -3614,7 +3614,7 @@ int WINAPI FarEditorControlA(int Command,void* Param)
 			if (Param)
 			{
 				oldfar::EditorSaveFile *oldsf = (oldfar::EditorSaveFile*) Param;
-				newsf.FileName=(oldsf->FileName)?AnsiToUnicode(oldsf->FileName):nullptr;
+				newsf.FileName=AnsiToUnicode(oldsf->FileName);
 				newsf.FileEOL=(oldsf->FileEOL)?AnsiToUnicode(oldsf->FileEOL):nullptr;
 			}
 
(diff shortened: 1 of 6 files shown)


far2l/vtshell.cpp
```diff
@@ -1019,7 +1019,7 @@ static bool shown_tip_exit = false;
 			return;
 	}
 	
-	~VTShell()
+	virtual ~VTShell()
```
elfmz: but why? its not really needed here - there're no descedants of VTShell, and  if make its d-tor virtual then better to same fot c-tors of interfaces that it derived from

far2l/wrap.cpp
```diff
@@ -3614,7 +3614,7 @@ int WINAPI FarEditorControlA(int Command,void* Param)
 			if (Param)
 			{
 				oldfar::EditorSaveFile *oldsf = (oldfar::EditorSaveFile*) Param;
-				newsf.FileName=(oldsf->FileName)?AnsiToUnicode(oldsf->FileName):nullptr;
+				newsf.FileName=AnsiToUnicode(oldsf->FileName);
```
elfmz: this doesnt look as correct at all

singalen: So you think these lines should stay the way they are?

elfmz: i think better to live without virtual of ~VTShell, or add empty virtual d-tor for all interfaces that it derived from.
Having virtual d-tor of class that derived from bases that have no virtual d-tor 'pushes' programmer to write buggy code, as he could assume that all.

singalen: Oh, shame on me, I completely missed the fact that it has base classes. Removing this "fix".

PR #452, opened 2018-12-18
Cleanup

svost:


far2l/tvar.cpp
```diff
@@ -436,6 +436,18 @@ int64_t TVar::getInteger() const
 	return ret;
 }
 
+int32_t TVar::getInt32() const
+{
```
elfmz: why not simple return static_cast<int32_t>getInteger()?

elfmz: ..or not _complicated_ int64_t ret = .. / ret = str ? atoi(str) : 0;  etc - so, not fully 'ported' to int32_t code?

PR #494, opened 2019-02-07
Incsrch

rampitec:
This is the port of my old plugin to search in the editor. I am using it now under Linux.
You either press ^I and start typing to search as you type or select a pattern and press ^Enter to search for it forward or ^#Enter to search backward.
Kudos for the Far port!


elfmz:
Russian language files must be converted to UTF8

rampitec:
> Russian language files must be converted to UTF8

They are converted, those which are installed by CMakeList.txt. These are the files ending with "u". There is a second set of files in CP866, these are used with Windows FAR plugin. The same source is used to build both Windows and Linux versions.

rampitec:
> > Russian language files must be converted to UTF8
> 
> They are converted, those which are installed by CMakeList.txt. These are the files ending with "u". There is a second set of files in CP866, these are used with Windows FAR plugin. The same source is used to build both Windows and Linux versions.

That actually makes sense. The root directory became a mess after the port. Moved it to the subdirectories. I still want to maintain a single source for Win and *nix, and I still do not want to duplicate English files under version control, so it is ENG, OEM and UTF8 rather than otherwise preferable Win and Nix.

Issue #342, opened 2018-06-18
Green autocompletion pops up in the editor and marks the file dirty


singalen:
I see it quite often.
I see not way to switch it off.
Even if I cancel the autocompletion, the editor still considers the file changed and asks for saving.

![screenshot 2018-06-18 12 10 19](https://user-images.githubusercontent.com/151199/41556833-a9d642b8-72f0-11e8-8c0b-fa767578c6e8.png)

volth:
to turn autocomplete off:
```
mkdir -p $HOME/.config/far2l/REG/HKU/c/k-Software/k-Far2/k-Plugins/k-editorcomp
echo -e 'Enabled\nDWORD\n0' > $HOME/.config/far2l/REG/HKU/c/k-Software/k-Far2/k-Plugins/k-editorcomp/v-Enabled
```

singalen:
Thanks! Why didn't I think about it. Just deleted the plugin.

Alter-1:
I've created patch for this issue. Undo (ECTL_UNDOREDO) should be used instead of suggested characters removal (ECTL_DELETECHAR)
[editorcomp-undo.patch.gz](https://github.com/elfmz/far2l/files/2989079/editorcomp-undo.patch.gz)

PR #503, opened 2019-03-21
fix for marking edited file dirty after cancelled autocomplete

Alter-1:
fix for https://github.com/elfmz/far2l/issues/342
Undo (ECTL_UNDOREDO) should be used instead of suggested characters removal (ECTL_DELETECHAR) when user cancels suggestion.
For proper operation suggestion is enclosed with EUR_BEGIN/EUR_END requests


Alter-1:
А почему ? Вроде бы pull request в эту ветку. Или я что-то как всегда пропустил ?

PR #512, opened 2019-03-26
fix for Hex pgUp/Down offset missmatch issue #474

Alter-1:
::Up() always uses byte offset, must be 16
::ReadString() uses byte count for 8/16-bit charsets, but word-count for 32-bit charset (checked by IsFullWideCodePage()). Must be 16 and adjusted for 32-bit charsets (those are not supported yet)


russiandesman:
I had a plan to investigate and fix it by myself, and what a pleasant surprise!
@elfmz any plans to merge this fix in nearest future?

PR #518, opened 2019-04-10
Fix build on FreeBSD 12

ZOleg73:


CMakeLists.txt
```diff
@@ -11,13 +11,18 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 if(NOT DEFINED CMAKE_INSTALL_PREFIX)
-   set(CMAKE_INSTALL_PREFIX "/usr")
+   set(CMAKE_INSTALL_PREFIX "/usr/local")
```
elfmz: Is it really neccessary? it will make update complications for those who is using default prefix

elfmz: ..so if its necessary for BSD build then wrap it into if (${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD") / else (old prefix) fi

ZOleg73:
Ok, I do it

PR #565, opened 2019-07-08
Feature/aliases resolving

russiandesman:
Can we merge this bash-only version?
In the future if someone needs this feature for other shells -- it is easy to add switch on getenv('SHELL') and obtain the same, the matter is only in `alias` output for particular shell.

### DIFF
--- a/far2l/execute_oscmd.cpp
+++ b/far2l/execute_oscmd.cpp
@@ -67,80 +67,24 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <wordexp.h>
 #include <sys/wait.h>
 
-int ipopen(FILE** stdout_handle, const char* command) {
-	int out[2];
-	int pid = -1;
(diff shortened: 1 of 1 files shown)


far2l/execute_oscmd.cpp
```diff
@@ -66,6 +66,26 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
+std::string resolve_alias(std::string str)
+{
+    std::string result;
+    FILE* stream;
+    const int max_buffer = 256;
+    char buffer[max_buffer];
+    char command[max_buffer];
+
+    sprintf(command, "bash -i -c \"alias %s | grep -Po \\\"(?<=alias %s=\').*(?=\')\\\"\"", str.c_str(), str.c_str());
```
elfmz: please use snprintf or (better) StrPrintf (from far2l utils)

russiandesman: done

PR #566, opened 2019-07-09
Add ability to specify initial path for panels from command line

iliakonnov:
Allows to run `far2l //bin //etc` to open `/bin` and `/etc` in two panels. Originally they were treated as flags (`/B` and `/E`) instead of pathes. Double slash is used to distinguish flags and UNIX path.

### DIFF
--- a/far2l/main.cpp
+++ b/far2l/main.cpp
@@ -79,7 +79,7 @@ static void CopyGlobalSettings();
 static void show_help()
 {
 	WCHAR HelpMsg[]=
-	    L"Usage: far [switches] [apath [ppath]]\n\n"
+	    L"Usage: far [switches] [[//]apath [[//]ppath]]\n\n"
 	    L"where\n"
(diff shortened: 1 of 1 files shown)


elfmz:
Not very 'nice'.. May be 

> far2l /left /bin  /right /tmp

 would be better?

iliakonnov:
I've change it to `far2l /cd /etc /cd /bin`.

Also PointToNameUNC (and CutToNameUNC possible) from pathmix.cpp is doing strange thing: `//etc/` returns empty string, but `/etc/` returns just `etc`. Then Far is trying to [change directory](https://github.com/elfmz/far2l/blob/75d908f2f1df3439ce99db714bec59f2bd981018/far2l/main.cpp#L268) for panel and fails without additional slash.
I do not know what PointToNameUNC is doing, so I can't fix it. Instead of it, I added dirty hack to bypass this issue: https://github.com/elfmz/far2l/blob/75d908f2f1df3439ce99db714bec59f2bd981018/far2l/main.cpp#L511

PR #569, opened 2019-07-15
Show a letter on the panels when hidden files aren't displayed.

ha3flt:
Since not showing hidden, system, etc. files is some kind of a filter itself, too, I wanted to see if this filter is active or not (e.g. toggled by Ctrl-H). If there is no hidden/system file(s) in a folder at all, the user cannot see the state of this switch. So, if the filter is active, an 'h' letter is shown next to the sort mode, and filter (asterisk) character(s) (if they are displayed).

Screenshot from 2019-07-15 19-25-20

I know, it is quite a change in the user interface having two or more characters in that position, but perhaps you'll like it. :-)


elfmz:
this PR looks little bit outdated so i cherry-picked its essential commit by https://github.com/elfmz/far2l/commit/e26057c238a902b9dcca397bbe0f2d74d8fb0a2b and thus closing this

ha3flt:
Hi,

It was my modification, thank you. Also for the great software/port, I 
use it everyday for everything. I hope I can contribute with something 
more valuable too in the future. :-)

Regards,
ha3flt

valdemar1977:
Using language independent symbol for indicate that hidden file is not showed imho more preferable. For example "#" symbol. Are you agree?

PR #571, opened 2019-07-28
Save isMaximized in consolesize

EduardSergeev:
Apart from console dimensions also save the fact if it is maximized
Maximize screen on startup if it was previously closed while being maximized


elfmz:
AFAIR it was discussed some time ago and concluded that its window manager's job :)

PR #586, opened 2019-09-03
use gio trash if gvfs-trash is not available

j2969719:
#548

### DIFF
--- a/far2l/bootstrap/trash.sh
+++ b/far2l/bootstrap/trash.sh
@@ -10,6 +10,10 @@ set -e
 
 if [ -x ~/.config/far2l/trash.sh ]; then
 . ~/.config/far2l/trash.sh
-else
-  gvfs-trash -f "$1" || gio trash -f "$1"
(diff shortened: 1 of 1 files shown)


far2l/bootstrap/trash.sh
```diff
@@ -10,6 +10,6 @@ set -e
 
 if [ -x ~/.config/far2l/trash.sh ]; then
 . ~/.config/far2l/trash.sh
+else
```
elfmz: Else is redundant (and even wrong )

far2l/bootstrap/trash.sh
```diff
@@ -10,6 +10,6 @@ set -e
...
+  gvfs-trash -f "$1" || gio trash -f "$1"
```
elfmz: using command -v would be 'nicer', see as example: https://github.com/elfmz/far2l/blob/master/far2l/bootstrap/notify.sh

PR #674, opened 2020-03-01
Syslog

sergpolkin:
In Debug mode build, log messages can be useful.
CMake variable SYSLOG enable/disable messages type.
Checked only on Linux machine.


elfmz:
you can get debug output right now by:
export FAR2L_STD=-
far2l

instead of - you can also specify log file name where it will be printed instead of stderr as with dash

Issue #502, opened 2019-03-13
Ukrainian I/i letter support for OEM cp866 charset


Alter-1:
Ukrainian I/i letter support for OEM cp866 charset.

There is well-known Windows problem with Ukrainian letters 'I' and 'i'. They are displayed as question mark ('?') in console applications. 
OEM cp866 characters 0xf6 (I ukr) and 0xf7 (i ukr) should be mapped to unicode 0x0406 and 0x0456 (http://alter.org.ua/en/soft/linux/wine_cp866u/)

The patch proposed adds support for ukrainian cp866 (cp866ukr) as codepage 22866
(like koi8r, koi8u) with 866ukr OEM-Unicode translation
[far2l_cp866uk.patch.gz](https://github.com/elfmz/far2l/files/2963715/far2l_cp866uk.patch.gz)

singalen:
Wow, there were multiple variants of 866. https://en.wikipedia.org/wiki/Code_page_866#Character_set
BTW, your website is down.

singalen:
I have submitted this patch as a PR.
Was there any reason to put the `cp866uk` where it was? I have moved it right after `cp866`, did it break anything?

singalen:
Can be closed, I believe?

PR #675, opened 2020-03-02
#502: Ukrainian I/i letter support for OEM cp866 charset

singalen:
Submitting #502 as a PR.


WinPort/wineguts/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ codepages/c_863.c
 codepages/c_864.c
 codepages/c_865.c
 codepages/c_866.c
+codepages/c_866.uk.c
```
elfmz: May be ua but not uk (thats United Kingdom)?

singalen:
I was surprised too, but it's weird. Ukraine is `UA` in [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements) and in Internet domains (British domains are .uk).
But [in POSIX locale](https://www.gnu.org/software/gettext/manual/html_node/Usual-Language-Codes.html#Usual-Language-Codes), which this code is about, Ukrainian language is `uk`.

elfmz:
funny but then ok

PR #677, opened 2020-03-03
Allow to use PRIMARY selection instead of CLIPBOARD X11 selection.

dstogov:


elfmz:
suggest in future to move it to UI settings but for now OK as command line

dstogov:
Thanks for your work!

I've been using FAR under Wine on Linux for almost 10 years.
But with new Wine versions it became less and less stable.
I finally, decided to try migrating to far2l. Usage of primary selection, for easy copy/paste between FAR editor, terminal and brawser, was one of the missing features, that stopped me from using far2l early.

It still has few annoying differences in behavior, but today was my first working day with far2l instead of FAR :)

PR #721, opened 2020-07-06
Updated "Ubuntu" to "Debian/Ubuntu" and addressed a broken link

randolf:
I use Far Manager successfully on Debian using the same instructions as on Ubuntu, so I added Debian to the documentation to reflect this.

Also added an archival link for CodeLite because the active link http://codelite.org/LiteEditor/WorkingWithCMake is no longer working and I couldn't find it on the current web site.


elfmz:
looks like correct link now is https://wiki.codelite.org/pmwiki.php/Main/TheCMakePlugin

PR #743, opened 2020-10-01
Fix detection of LibArchive library on MacOS

anteo:
In the latest toolchain LibArchive is included, so even if brew has own version of LibArchive, the version from SDK is used causing build errors. Need to set library location explicitly.


anteo:
Sorry, please revert this one.
I found that libarchive is not used at all without a call to **find_package(LibArchive)**
I'll provide a fix later.

It now looks like latest libarchive has 7Z support so it starts interfering with 7z sources in 7z/C

PR #744, opened 2020-10-01
Fix the issue when FAR2L is not rebuilt automatically when WinPort sources are changed

anteo:
I noticed that when I change WinPort library sources, main executable is not rebuilt automatically.
I added dependency on libWinPort.a, so when library object file is changed, main executable is also rebuilt.

### DIFF
--- a/far2l/CMakeLists.txt
+++ b/far2l/CMakeLists.txt
@@ -164,7 +164,7 @@ target_compile_definitions(far2l PRIVATE -DUNICODE)
 target_include_directories(far2l PRIVATE Include)
 target_include_directories(far2l PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
-add_dependencies(far2l bootstrap)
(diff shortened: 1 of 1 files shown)


far2l/CMakeLists.txt
```diff
@@ -173,6 +173,11 @@ else()
     SET (WINPORT -Wl,--whole-archive WinPort -Wl,--no-whole-archive)
 endif()
```
elfmz, line 175: i think below lines are redundant, add_dependencies must be enough

anteo: Unfortunately no,
You can try by yourself. I've just  reproduced this again, by commenting out **set_source_file_properties** section below. Main file doesn't recompile.
I'm not familiar with Cmake internals, just found similar issue:
https://cmake.org/pipermail/cmake/2014-October/058791.html
That was why I implemented fix this way

PR #745, opened 2020-10-01
Fix detection of output of commands executed in non-bash shells

anteo:
Current implementation skips only one line of output when receives output of shell command.
Which in standard case contains "magic" call with $FARVTRESULT and completion marker.
But in some shells like ZSH with OhMyZsh installed, the output of command may contain a lot of ANSI sequences and have more than one line, causing this "magic" call to appear  in console. 

Idea is to add output "starting" marker, additionally to completion marker, so parser could ignore everything up to starting marker and then copy output until completion marker.

Screenshot 2020-10-01 в 21 40 40

### DIFF
--- a/far2l/vtshell.cpp
+++ b/far2l/vtshell.cpp
@@ -344,94 +344,170 @@ private:
 	}
 };
 
-class CompletionMarker 
+class Marker
 {
-	std::string _marker;
-	std::string _backlog;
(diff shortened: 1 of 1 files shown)


elfmz:
i have constantly deferred idea to replace this marker with specific ESC-sequence, but ok for now, while that idea still deferred)

PR #747, opened 2020-10-01
Fix rendering of fonts with fractional width

anteo:
E.g. if font size = 13, one char in a string has width 9px (getExtent() returns 9), but total string width is less than N*9px. It causes rendering artifacts when render buffering is enabled. Idea is to disable buffering for certain font sizes as it is done for non-monospaced fonts:

Screenshot 2020-10-01 в 21 46 38

I haven't checked this issue in Linux but in MacOS it surely has a place.
I found "nobufferring" option first of all and enabled it, which fixed this issue, but I think this case could be easily detected in the code, which is my PR is all about.

### DIFF
--- a/WinPort/src/Backend/WX/Paint.cpp
+++ b/WinPort/src/Backend/WX/Paint.cpp
@@ -173,8 +173,15 @@ class FontSizeInspector
 			if (_prev_height!=-1) _unstable_size = true;
 			_prev_height = height;
 		}		
-	}	
-	
+	}
+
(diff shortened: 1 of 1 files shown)


elfmz:
there're no such issue on Ubuntu, so suggesting to make it osx-only ( #ifdef __WXOSX__ )

anteo:
> there're no such issue on Ubuntu, so suggesting to make it osx-only ( #ifdef **WXOSX** )

Sure

PR #750, opened 2020-10-02
Colorer update

elfmz:
Colorer: update to recent colorer-library and schemes DB


elfmz:
touch #692, close #648, touch #596, touch #531, touch #286

Issue #720, opened 2020-06-23
strange path changes with multiarc


unxed:
1. run fresh instance of far2l (it's important!)
2. open any archive on right panel
3. press Alt+F1
4. press Enter
5. press Tab
6. press Enter
7. left panel folder is changed to "/usr/lib/far2l/Plugins/multiarc/plug/ "

o_O

unxed:
The problem is here:
https://github.com/elfmz/far2l/blob/master/far2l/PluginA.cpp

Uncommenting all commented lines from 305 to 325 fixes it.

```
diff --git a/far2l/PluginA.cpp b/far2l/PluginA.cpp
index 9e0e6add..2dfab687 100644
--- a/far2l/PluginA.cpp
+++ b/far2l/PluginA.cpp
@@ -303,7 +303,7 @@ bool PluginA::Load()
 		return true;
 
 	if (!m_hModule)
-	{/*
+	{
 		FARString strCurPath, strCurPlugDiskPath;
 		wchar_t Drive[]={0,L' ',L':',0}; // ставим 0, как признак того, что вертать обратно ненадо!
 		apiGetCurrentDirectory(strCurPath);
@@ -314,14 +314,14 @@ bool PluginA::Load()
 			Drive[1] = m_strModuleName.At(0);
 			apiGetEnvironmentVariable(Drive,strCurPlugDiskPath);
 		}
-*/
+
 		PrepareModulePath(m_strModuleName);
 		m_hModule = WINPORT(LoadLibraryEx)(m_strModuleName,nullptr,LOAD_WITH_ALTERED_SEARCH_PATH);
 		GuardLastError Err;
-		//FarChDir(strCurPath);
+		FarChDir(strCurPath);
 
-	//	if (Drive[0]) // вернем ее (переменную окружения) обратно
-		//	WINPORT(SetEnvironmentVariable)(Drive,strCurPlugDiskPath);
+		if (Drive[0]) // вернем ее (переменную окружения) обратно
+			WINPORT(SetEnvironmentVariable)(Drive,strCurPlugDiskPath);
 	}
 
 	if (!m_hModule)
```

PR #752, opened 2020-10-03
Remove unneeded comments (fixes https://github.com/elfmz/far2l/issues…

unxed:
Fix for https://github.com/elfmz/far2l/issues/720


elfmz:
sorry but i'v commited fix that bit different to your PR (there is code really not needed - per-drive environment variables irrelevant for far2l, needed only to restore curdir)

unxed:
It's ok, thanks!

Issue #753, opened 2020-10-03
multiarc fails to extract .zip files generated by google forms if russian chars are used in form name


unxed:
Sample:

[csv.zip](https://github.com/elfmz/far2l/files/5321919/csv.zip)

unxed:
Workaround:
```
diff --git a/multiarc/src/arccmd.cpp b/multiarc/src/arccmd.cpp
index 040e5ece..9889c934 100644
--- a/multiarc/src/arccmd.cpp
+++ b/multiarc/src/arccmd.cpp
@@ -109,6 +109,19 @@ bool ArcCommand::ProcessCommand(std::string FormatString, int CommandType, int I
     Hide = 0;
 
   ExecCode = Execute(this, Command, Hide, Silent, NeedSudo, Password.empty(), ListFileName);
+
+  // charset workaround for unzip
+  if (strncmp(Command.c_str(), "unzip ", 6) == 0) {
+    if (ExecCode == 1) {
+      // "1" exit code for unzip is warning only, no need to bother user
+      ExecCode = 0;
+    } else if (ExecCode == 11) {
+      // "11" means file was not found in archive. Maybe unzip guessed charset wrong, retry as utf8
+      Command.replace(0, 6, "unzip -I utf8 -O utf8 ", 0, 22);
+      ExecCode = Execute(this, Command, Hide, Silent, NeedSudo, Password.empty(), ListFileName);
+    }
+  }
+
   if (ExecCode==RETEXEC_ARCNOTFOUND)
     return false;
```

As unzip's internal charset detection logic differs from far2l's one, to fix this properly we should remember if CPToUTF8() was called in ZIP_GetArcItem(), and, if so, append "-I CHARSET -O CHARSET" to unzip arguments. If CPToUTF8() was not called, we should append "-I utf8 -O utf8" to unzip arguments. Probably we should introduce a new %% variable for that.

And to work properly with non-russian charsets, we should not assume CP_OEMCP is always 866, but [detect it from system locale](https://github.com/unxed/oemcp/blob/master/oemcp.txt) just as windows does. Same thing with CP_ACP, but such archives are rare and almost never found in the wild.

elfmz:
workarond ok for now as workaround. but better use libarchive then, but need to add password support to related code (and hope libarchive supports password'ed zips, unlike 7z-s..)

unxed:
[...]

PR #754, opened 2020-10-03
Multiarc utf8 workaround

unxed:
Workaround for problem with multiarc and .zip archives generated by Google Forms. See https://github.com/elfmz/far2l/issues/753


elfmz:
dirty, but..

unxed:
> dirty

Better solution is proposed in #753 ;-)

PR #755, opened 2020-10-03
Smarter processing of non-latin filenames inside .zip archives

unxed:
It's better to select code page for OEM .zip archives based on system locale instead of always using code page 866.

That is exectly how Windows internal "zip folders" tool works. Locale-to-CP translation table taken from Wine sources.
See https://github.com/elfmz/far2l/issues/753 for details

### DIFF
--- a/multiarc/src/MultiArc.cpp
+++ b/multiarc/src/MultiArc.cpp
@@ -73,6 +73,9 @@ SHAREDSYMBOL void WINAPI _export SetStartupInfo(const struct PluginStartupInfo *
 
 SHAREDSYMBOL HANDLE WINAPI _export OpenFilePlugin(const char *Name,const unsigned char *Data,int DataSize,int OpMode)
 {
+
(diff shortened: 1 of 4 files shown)


unxed:
In addition to supporting locales other than Cyrillic, this code ensures that unzip will be called using exactly the same encoding as used by multiarc to interpret file list: if not on the first try, then on the second.

multiarc/src/MultiArc.cpp
```diff
@@ -299,3 +303,86 @@ SHAREDSYMBOL void WINPORT_DllStartup(const char *path)
...
+
+int oem_cp;
+
+void detect_oem_cp() {
+
+  // detect oem cp from system locale
+
+  oem_cp = 437; // default
+
+  char *lc = setlocale(LC_CTYPE, "");
```
elfmz, line 315: as i understand second arg of it setlocale must be NULL, but not empty string

PR #766, opened 2020-10-05
[multiarc] no more unneeded zip/unzip retries; ANSI support; fixed interpreting wrongly transcoded chars ("?"s) as wildcards

unxed:


unxed:
@elfmz ?

unxed:
At least this fixes deleting file from this archive 
[new2.zip](https://github.com/elfmz/far2l/files/5340668/new2.zip)

unxed:
This also fixes a problem then deletion of "абвгде" file from [this archive](https://github.com/elfmz/far2l/files/5342256/winzip.zip) forces deletion of "жзийкл" file also when working under el_GR locale.

unxed:
Eliminated the need for zip/unzip call retries: correct charset is now specified at first try.

Tested on my .zip test set, working as a charm.

PR #767, opened 2020-10-06
Workaround for Unzip to work in MacOS (skips recent OEM encoding checks for now)

anteo:
This is related to work you guys did recently.

Unzip in MacOS doesn't have -I or -O options, so extracting ANY files just doesn't work.

I have no idea how it could be solved without -I and -O, so let's skip new logic for MacOS (as a quick workaround)

### DIFF
--- a/multiarc/src/arccmd.cpp
+++ b/multiarc/src/arccmd.cpp
@@ -108,6 +108,12 @@ bool ArcCommand::ProcessCommand(std::string FormatString, int CommandType, int I
   if ((Hide == 1 && CommandType == 0) || CommandType == 2)
     Hide = 0;
 
+// Unzip in MacOS doesn't have -I and -O options thus it fails when extracting any file,
+// Would need another workaround, but let's just skip it for now
+#ifdef __WXOSX__
+  ExecCode = Execute(this, Command, Hide, Silent, NeedSudo, Password.empty(), ListFileName);
+  fprintf(stderr, "ArcCommand::ProcessCommand: ExecCode=%d for '%s'\n", ExecCode, Command.c_str());
+#else
   // charset workarounds for unzip
   if (strncmp(Command.c_str(), "unzip ", 6) == 0) {
     // trying as utf8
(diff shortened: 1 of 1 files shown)


unxed:
> I have no idea how it could be solved without -I and -O, so let's skip new logic for MacOS (as a quick workaround)

On MacOS we may try to transcode unzip parameters just as we do it with "zip -d" calls.

anteo:
Libarchive solution looks more generic to me.

What is currently done looks as hack anyway, forcing ProcessCommand (common function for all formats) to check substring 'unzip' or 'zip' in the command and injecting some logic. ProcessCommand should not know anything about format that was used, its job just to execute command; So maybe "retry" logic should be implemented in specific format?

Also is it related to ZIP only or any can be related to other formats as well? If so, logic should be generalized.
Thus moving charset handling entirely to Libarchive could be more preferable.   

I'd suggest this PR to be merged for now, because it's not possible to work with ZIP at all on MacOS.

unxed:
>  "retry" logic should be implemented in specific format

Already done that https://github.com/unxed/oemcp/blob/master/zipwrapper

elfmz:
Unfortunately libarchive of some versions (like that comes with ubuntu 20) crashes on such archives, and some older version (like in ubuntu 16) just fails to parse such archive.
May be right way is to use libminizip thats already linked with colorer. But first need to check how it will handle such zip files.
Transcoding shim also looks better workaround than hardcpded command line hacks. It can be multiarc itself, via special export

unxed:
>  It can be multiarc itself, via special export

Of course I tried if I can write the same shim around multiarc's libarchive for .zips :) No luck though

Also if unzip has no charset options like -I/-O, the only possible workaround is to rename extracted file[s] trying different charsets. Possible, but a bit weird.

unxed:
Found zip charset solution acceptable for MacOS: https://github.com/elfmz/far2l/issues/775

unxed:
Possible alternative to unzip under MacOS is using unar. Implemented here https://github.com/elfmz/far2l/pull/766

Issue #786, opened 2020-10-14
far2l's internal VT sends some key combinations incorrectly


unxed:
Steps to reproduce:

1. sudo apt install [tilde](https://github.com/gphalkes/tilde)
2. tilde somefile.txt
3. Try Shift+Home/End or Ctrl[+Shift]+Home/End in far2l's VTE, compare to GNOME Terminal

unxed:
Affects tilde also (sudo apt install tilde).

unxed:
Fix proposed https://github.com/elfmz/far2l/pull/787

unxed:
finally fixed in https://github.com/elfmz/far2l/commit/fde42d9566acc7099f240ab05120b7dd4d384c47

the first proposed solution was to reset keypad var to 0 for home/end keys if ctrl or alt or shift flag present (just leave it here for historical reasons)

PR #787, opened 2020-10-14
fix for (ctrl|shift)+(home|end) keys combinations behavior...

unxed:
...in mc's internal editor and tilde TUI editor. see https://github.com/elfmz/far2l/issues/786


unxed:
finally fixed in https://github.com/elfmz/far2l/commit/fde42d9566acc7099f240ab05120b7dd4d384c47

Issue #796, opened 2020-10-16
Opening editor terminates far2l


cwollenhaupt:
I built the current version using make, because the one line install failed for me. far2l works fine when I launch it. I can navigate and view files with F3. However, every time I open the editor, either by pressing F4 in the panel view or F6 from the file view window, then far2l simply disappears.

strim:
Same with my far2l build for Mac OS X Catalina 10.15.7

cwollenhaupt:
That's the same macOS version I'm using. I should have added that in the first place.

strim:
I checked the latest master, editor crashes as you described. Attaching [crash.log](https://github.com/elfmz/far2l/files/5393982/crash.log). As I can interpret, the culprit is colorer.

singalen:
I sent a pull request: #799.

strim:
Thanks a lot, I confirm the issue is fixed in my build with the latest master

PR #799, opened 2020-10-16
#796: Opening editor terminates far2l (symbol not found). Couple of warnings.

singalen:
…arnings fixed.

### DIFF
--- a/colorer/src/Colorer-library/src/colorer/common/Exception.cpp
+++ b/colorer/src/Colorer-library/src/colorer/common/Exception.cpp
@@ -25,7 +25,7 @@ Exception &Exception::operator=(const Exception &e) noexcept
   return *this;
 }
 
-Exception::~Exception()
+Exception::~Exception() noexcept
 {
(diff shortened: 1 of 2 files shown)


singalen:
People also use `__unix__` paired with `__APPLE__`:
* https://stackoverflow.com/questions/5919996/how-to-detect-reliably-mac-os-x-ios-linux-windows-in-c-preprocessor, 
* https://stackoverflow.com/questions/2989810/which-cross-platform-preprocessor-defines-win32-or-win32-or-win32
I'll check out the CMakeLists, maybe I can fix those.

anteo:
it is because of this commit:
https://github.com/elfmz/far2l/commit/5d798ef7173d19cc6339834ae83b3868442a41ce#diff-43a412227ddcdb64dcc66ba3055934b0f1dc8d5bbceff1c14066b5c1405283a2L74

which removed __unix__ from target_compile_definitions of colorer.

But this solution works too

PR #808, opened 2020-10-20
remove non-working items from UI

unxed:
https://github.com/elfmz/far2l/issues/84#issuecomment-712912910


far2l/plist.cpp
```diff
@@ -48,6 +48,9 @@ static BOOL KillProcess(DWORD dwPID);
 
 void ShowProcessList()
 {
+    // Currently unsupported on linux
+    return;
```
elfmz, line 52: why return here? its not called anyway

unxed: called by Ctrl+W

unxed: > why return here? its not called anyway

still callable by Ctrl+W

elfmz: may be just run htop?

elfmz: or user script (like roots.sh) that will run htop if it installed, otherwise - top..

unxed: > or user script (like roots.sh) that will run htop if it installed, otherwise - top..

the problem is Ctrl+W can be pressed in viewer/editor also

what should we do in those cases?

elfmz:
- it experimented a bit with htop launching by  ctrl+w works inside editor & viewer but after exit from htop there re some visual artifacts remaining, must be fixable by kind of redraw, will complete experiments later

elfmz:
so, i've commited using of htop/top, plz leave only MMenuWindowsList removal in your PR

unxed:
оставил отключение только пункта про hotplug

unxed:
> so, i've commited using of htop/top, plz leave only MMenuWindowsList removal in your PR

done, still ps.sh is missing in your process list update commit

PR #834, opened 2020-10-30
fix for free pascal's tui ide interface

unxed:
Fix for #830


unxed:
NDN's UI is also fixed by this patch

elfmz:
This is DEC Special Graphics
https://en.wikipedia.org/wiki/DEC_Special_Graphics
http://fileformats.archiveteam.org/wiki/DEC_Special_Graphics_Character_Set
why they enable it when printing text, but not graphics?

PR #852, opened 2020-12-04
there is no "nproc" on mac, use "sysctl -n hw.logicalcpu" instead

trexinc:


elfmz:
ok for now, but may consider to use 'getconf _NPROCESSORS_ONLN' as common denominator that works both in linux and macos

Issue #797, opened 2020-10-16
How do I debug libarchive problems?


singalen:
In order to start diagnosing it, I could use a hint. What is going on here?
![Screen Shot 2020-10-16 at 13 07 09](https://user-images.githubusercontent.com/151199/96304275-aad9e700-0fb0-11eb-87db-38d65fa21a0f.png)

elfmz:
what that strange archives BTW?

elfmz:
BTW remembered that there were problems with libarchive (or something else??), may be even form you. And it appeared that detected headers where mismatched to actual library binary linked to.
Something came from brew, something from OS

singalen:
True, that was me. I have one version from Apple, for which headers are not available and which is on the default library path, and a version from MacPorts which has headers and thus gets used for the build.
I need to invent something to link with the MacPorts version in runtime, probably add it to the installed binaries and add a RPATH.

elfmz:
its should be now possible to build far2l package without brews etc with all dependencies: https://github.com/elfmz/far2l/pull/764
I didnt test though, i'm not using mac.

singalen:
Yep, that's excellent. I need to test how that mechanism behaves in the described case of conflicting libraries. Probably badly and that's what I'm observing.

elfmz:
Ping here, too. Check if problem still there

singalen:
Somehow it works now.

I honestly don't understand why, because multiarc is linked with a system version:
```
~/src/amusement/far2l$ otool -L cmake-build-release/install/far2l.app/Contents/MacOS/Plugins/multiarc/plug/multiarc.far-plug-mb 
cmake-build-release/install/far2l.app/Contents/MacOS/Plugins/multiarc/plug/multiarc.far-plug-mb:
        /usr/lib/libpcre.0.dylib (compatibility version 1.0.0, current version 1.1.0)
        /usr/lib/libarchive.2.dylib (compatibility version 9.0.0, current version 9.2.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```
but the only headers I have is in MacPorts version, in `/opt/local`, and the headers are taken from `/opt/local/include`, I logged it in `multiarc/CMakeLists.txt`.

elfmz:
[...]

PR #859, opened 2020-12-14
#797: workaround for macports' problem with libarchive.

singalen:
Copied a brew's libarchive hack for MacPorts.


elfmz:
but what about LibArchive_INCLUDE_DIR  ?

singalen:
System `libarchive` comes without headers, that's the whole problem. Otherwise, we would happily link with the system copy and never use one from macports/homebrew.

PR #876, opened 2021-01-08
Add readme for NixOS builds, and how to update package.

hypersw:
`unxed` said I should probably write this down somewhere. This looks like the place.


elfmz:
BTW Travis CI config for nixos build would be nice too..

Issue #909, opened 2021-02-13
Copy to clipboard full path or short name of current file in a panel, Mac OS X


strim:
Windows Far2 used to copy name of the currently highlighted file in a panel to clipboard with shortcuts Ctrl+Ins or Ctrl+C. Same was for any text entered by user to command-line prompt under the panels. This way I could do, for example, Ctrl+F to paste full path of the current panel file to the prompt, and next to grab it with Ctrl+Ins to clipboard. That was great to get full path from Far and to go to other apps with it. 

Far2l at Mac OS X is able to paste to under-panel prompt full file path and short file name with Ctrl+F and Ctrl+Enter. However, I found no hotkey to grab the prompt content to Mac clipboard. Neither Cmd+C, nor Ctrl+C works here. Is there any way to reach my goal? Thank you,

elfmz:
Ctrl+Ins works on my linux system, but right now i can't what happens on macos. Also try Cmd+Ins.

strim:
Thank you for checking on Linux. Since neither Apple laptops, nor Apple keyboards have Insert key, there is no way to try the shortcut as suggested. I guess even for PC keyboard connected to Mac OS it won't support any meaningful behavior for Ctrl+Ins or Win+Ins combinations.

singalen:
There's a #575 reporting the same problem, but it was closed – wrongfully, I believe. The "it's the original Far behavior" is not a good motivation to a missing feature.

My suggestion: Use `Ctrl-C` to break a process where there is a running process; otherwise, make it copy the current file/selection path(s). On Mac, it's even simpler, `Cmd-C` should work.

I'll see if I have the time to make a PR.

strim:
Thank you for the reference to the same issues reported. Now at least I know that switching panels off with Cmd+O helps to copy the prompt content with Cmd+C. Not ideal but gives the way to grab file path. Thank you!

singalen:
Couldn't quickly figure out how to distinguish Cmd from Ctrl events 8(
@elfmz, may I ask for some help? 
I see that one part would be to add a `case KEY_CTRLC` to `far2l/edit.cpp:1356`, but how do I distinguish Ctrl from Cmd?

elfmz:
Currently they're equivalent, why you want to distinguish? [Ctrl|Cmd]+C should work fine as desribed above

singalen:
If they are, then probably there's no reason to, thanks! Then I will continue working on my patch.

singalen:
[...]

PR #922, opened 2021-02-24
#909: Copy to clipboard the full path or short name of a current file on small keyboards.

singalen:
Added Cmd/Ctrl-C shortcut to copy current command or file(s), and Alt-Shift-C to copy full path(s).
Not sure this is all that needs to be done, please consider this WIP.

* Is there a plan to resurrect copying files for Explorer/Finder?
* This probably needs to be documented?
* Do we care for UNC paths?

### DIFF
--- a/far2l/filelist.cpp
+++ b/far2l/filelist.cpp
@@ -1086,22 +1086,27 @@ int FileList::ProcessKey(int Key)
 			Redraw();
 			return TRUE;
 		case KEY_CTRLINS:      case KEY_CTRLNUMPAD0:
-
-			if (CmdLength>0)
+		case KEY_CTRLSHIFTINS: case KEY_CTRLSHIFTNUMPAD0: case KEY_CTRLC:     // копировать имена
+		case KEY_CTRLALTINS:   case KEY_CTRLALTNUMPAD0:                       // копировать UNC-имена
+		case KEY_ALTSHIFTINS:  case KEY_ALTSHIFTNUMPAD0:  case KEY_ALTSHIFTC: // копировать полные имена
+			if ((Key == KEY_CTRLC || Key == KEY_CTRLNUMPAD0 || Key == KEY_CTRLINS) && (CmdLength > 0)) {
 				return FALSE;
-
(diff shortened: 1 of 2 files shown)


far2l/edit.cpp
```diff
@@ -1351,7 +1351,9 @@ int Edit::ProcessKey(int Key)
 			Show();
 			return TRUE;
 		}
-		case KEY_CTRLINS:     case KEY_CTRLNUMPAD0:
+		case KEY_CTRLINS:
```
elfmz, line 1354: are you sure its needed? There is Key reassignment for KEY_CTRLC: https://github.com/elfmz/far2l/blob/master/far2l/edit.cpp#L695

far2l/filelist.cpp
```diff
@@ -1086,22 +1086,26 @@ int FileList::ProcessKey(int Key)
 			Redraw();
 			return TRUE;
 		case KEY_CTRLINS:      case KEY_CTRLNUMPAD0:
-
-			if (CmdLength>0)
+		case KEY_CTRLSHIFTINS: case KEY_CTRLSHIFTNUMPAD0: case KEY_CTRLC:     // копировать имена
+		case KEY_CTRLALTINS:   case KEY_CTRLALTNUMPAD0:                       // копировать UNC-имена
+		case KEY_ALTSHIFTINS:  case KEY_ALTSHIFTNUMPAD0:  case KEY_ALTSHIFTC: // копировать полные имена
+		    if (CmdLength > 0)
```
elfmz, line 1092: why if (CmdLength > 0) moved?

far2l/edit.cpp
singalen: I am not, will check.

far2l/filelist.cpp
singalen: To unify the logic: if there is anything in a command line, copy it, and not the selected files. 
OTOH, not all of the shortcuts copy the command line... looks like we need more conditions here.

singalen:
Updated. Not the cleanest patch, but I guess it didn't make it worse.

Issue #698, opened 2020-05-13
Mac: Copying to clipboard stopped working


singalen:
This may have something to do with MacOS 10.14.6 update, or wxWidgets, or something, but Cmd-C no longer copies text to clipboard.

`wxClipboardBackend::OnClipboardClose()` is called, then the call goes into `wxTheClipboard->SetData()`, but nothing appears in the clipboard.

Any other Mac users with such a problem?

elfmz:
stupid question, but did you try reboot? i saw some strange issues on macos fixed by reboot, but not related to clipboard..

singalen:
Yes, I did. It didn't help 8(
Other hotkeys stopped working for me before, funny - this only happened when there was a system update pending (installed but not rebooted). But this is a different case, far receives a keystroke but wx doesn't connect the text to the system clipboard.

elfmz:
also does it work in reverse direction - pasting works?

singalen:
Pasting works.

elfmz:
i created branch with some extra logging: https://github.com/elfmz/far2l/tree/debug-clipbrd
build and run it like this:
FAR2L_STD=- ./install/far2l 2>&1 | grep -i clipboard

singalen:
Thank you very much.

```
OnKeyDown: 37 0 134 0 590972481
13:36:31: Debug: Unrecognised keycode 364
13:36:31: Debug: Unrecognised keycode 365
ConsoleInput::Enqueue: # 0 11 8 DOWN
OnKeyDown: 8 43 43 0 590973097
13:36:31: Debug: Unrecognised keycode 364
13:36:31: Debug: Unrecognised keycode 365
OnChar: 43 43 0 590973097 _lk_ts=590973097 _lk_enqueued=0
13:36:31: Debug: Unrecognised keycode 364
13:36:31: Debug: Unrecognised keycode 365
ConsoleInput::Enqueue: C 43 43 8 DOWN
ConsoleInput::Enqueue: C 43 43 8 UP
OpenClipboard
OnClipboardEmpty: dispatching into main thread...
EmptyClipboard: Clear
SetClipboardData: dispatching into main thread...
SetClipboardData: 13
OnClipboardClose: dispatching into main thread...
OnClipboardClose
OnClipboardClose: SetData result=1 prim_selection=0
OnClipboardClose: Flush result=0
```
`wxTheClipboard->SetData()` returned `true`. Not sure what primary selection means.

elfmz:
[...]

PR #942, opened 2021-02-28
#698: pasteboard hack.

singalen:
I got tired of rebasing my hack on top of master, and finally made another one, a bit less horrible.
Still will not help if `far2l` copies anything but text into a clipboard (I think?..)


singalen:
The error on FreeBSD is `Signaled to exit (terminated)!`.
I see no other problems there.

singalen:
Oh no, that's me being silly. Sure I should check for Mac too.
Updated the PR.

Issue #901, opened 2021-01-29
consider adding higher color-resolution support and using chafa instead of jp2a


unxed:
chafa gives much better results then jp2a. i'm using

chafa -c 16 --color-space=din99d --dither=ordered -w 9 --symbols all --fill all !.! && read -n1 -r -p ""

for "previewing" jpegs

unxed:
Source image:
![изображение](https://user-images.githubusercontent.com/1151423/106360934-3d1db500-632c-11eb-9abd-4841bfc7f602.png)

jp2a:
![изображение](https://user-images.githubusercontent.com/1151423/106360941-43ac2c80-632c-11eb-9c09-ebe19d5d378d.png)

chafa in far2l:
![изображение](https://user-images.githubusercontent.com/1151423/106360949-4e66c180-632c-11eb-95ed-b2f907baf56a.png)

chafa in true color gnome terminal:
![изображение](https://user-images.githubusercontent.com/1151423/106360955-59215680-632c-11eb-9997-31e06438bc30.png)

elfmz:
1st good reason to implement higher color-resolution

sxored:
@unxed can you check #966 commit with chafa ?

unxed:
> @unxed can you check #966 commit with chafa ?

Pretty cool commit! Already using it :)

unxed:
[str_visual] now then truecolor is supported we should switch to "chafa -c full" for better UX

elfmz:
beside chafa there is also timg

unxed:
anyway we should remove `-c 16` from `chafa -c 16` in view.sh

PR #966, opened 2021-03-27
View.sh update with more file types

sxored:
Added more file types to process by view.sh in internal viewer


sxored:
chafa was proposed by @unxed in #901

PR #973, opened 2021-04-05
View.sh update with various fixes

sxored:
Fixed view of some file types to process by view.sh in internal viewer


sxored:
continuation of #972

PR #975, opened 2021-04-07
debugging and dynamic python plugins

m32:


elfmz:
i dont think its right idea to commit vscode project, especially as a part of files that gonna be copied to user's system during make install

m32:
Unfortunately, I agree, but I had no idea where to copy these files to be also available without installing far2l sources.
Do you have any suggestion?

BTW How to install files in ~ .config/far2l/plugins/python?

elfmz:
just dont commit them

elfmz:
or put them inside of plugins sources dir with instructions how to use them (or with shell script that will copy them where necessary)

PR #999, opened 2021-05-08
Add GitHub Action for fresh builds

techtonik:
This packs build artifacts and makes them available for a limited time from Actions tab. I packed minimal with the intention to compare size with different options, such as Python, but didn't get to it.


elfmz:
but why dependencies.txt so short? NetRocks will not have even SSH support with such list

techtonik:
https://github.com/techtonik/far2l/actions/runs/858609936 - SSH and `libarchive` add only 300 kB packed. =) I am trying to make a Python build next.

techtonik:
@elfmz Python plugin failed to build with a strange message.

```
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython3.8.so (found suitable version "3.8.5", minimum required is "3.0") 
CMake Warning at CMakeLists.txt:410 (message):
-- SIMPLEINDENT plugin enabled
  PYTHON plugin disabled due to virtualenv or python3 not found
```

I am not proficient in CMake, but why https://cmake.org/cmake/help/latest/module/FindPython3.html is not being used?

elfmz:
regarding pythin, its a question to @m32

m32:
Nor am I an expert on cmake. My system (ubuntu 18.04 lts) has cmake 3.10.2 and I don't have this problem.
Maybe the libpython3-dev or virtualenv (?) packages are missing from the system (?), or the newer cmake has another way to find python libraries ...
I have no tip that can help

techtonik:
The error message clearly says that there is no `virtualenv`. I must have been blind. =)

https://github.com/elfmz/far2l/blob/1ea471e78c5792f4b67f64cc8839ec2d146f6570/CMakeLists.txt#L404

PR #1005, opened 2021-06-02
fix optimized clang build, fix static linking of libc++

pg83:


elfmz:
hm, can you describe what was the reason for that changes? they look little bit strange..

pg83:
> hm, can you describe what was the reason for that changes? they look little bit strange..

1. We have some places, where we call hasParent() with this == 0. It is forbidden by C++ standart, and compiler optimizes code based on assumption that this never equals zero.

2. Can not use std::cerr in loadable module, iff links with libc++ statically(my case), because we have two copies of static data.

3. We should propagate exception from xerces-c initialization, and does not use improperly initialized library.

Issue #1004, opened 2021-05-29
CPack uses invalid characters for the DEB package version number


Devoter:
I built a program from `master` branch and created a .deb package via cpack, but cannot install it. Dpkg throws an error:

```
'Version' field value '2.3.210528_2e3f6fc0_alpha': invalid character in version number
```

I guess, cpack uses something like `git describe --tags --long --always --dirty`, but it is a wrong way.

daemonserj:
Та же фигня :(

elfmz:
There was really git hash etc, now i replaced that with.. gits commit count. Since its  number that typically monotonically grows up, so should work fine as unique version number for lazy developers :) Please confirm if its fixed now for you.

Devoter:
I guess, the issue can be closed. Thanks for fixing my fix again.

PR #1011, opened 2021-06-17
Make the version compliant with the debian packaging rules

Devoter:
Solution of #1004


elfmz:
in PR was unnoticed mistake fixed by https://github.com/elfmz/far2l/commit/5b481af453fc453fad7342c1a28418f5829fe3d9 - check that your issue is really fixed afterwards

Devoter:
Oh, thanks! Sorry I was blind.

PR #1060, opened 2021-08-17
Updated ConvertUTF.c to fix licensing compatibility with Debian

unxed:
From https://github.com/bro/bro/blob/master/src/ConvertUTF.c

Added CalcSpace* functions from https://github.com/elfmz/far2l


alexmyczko:
tried this #4 at https://mentors.debian.net/package/far2l/ but it seems not good enough. not sure if it's really not good or the check is just bad. waiting the in the issue for an alternative solution, non-embedded source, with addition build-depends: is preferred

elfmz:
may be https://github.com/elfmz/far2l/pull/1061 will be better?

PR #1083, opened 2021-09-26
Fix includes

michaellukashov:
add missing includes used in modules


utils/src/InMy.cpp
```diff
@@ -5,6 +5,7 @@
 #include <unistd.h>
 #include <pwd.h>
 #include <assert.h>
+#include <mutex>
```
elfmz, line 8: but there is no mutex (anymore)

elfmz:
multiarc/src/formats/rar/unrar is 3rd party (and updated as whole from time to time) so don't do any changes there unless critically needed

PR #1130, opened 2021-12-05
update view.sh - added more file types

sxored:


elfmz:
uh, this small script becomes terrific :)

PR #1204, opened 2022-01-21
Trying different keyboard layout[s] on FastFind fail

unxed:
Solution for #780. Adopted for current master

Layout[s] translation table[s] configurable via .ini file, locale-dependent


unxed:
Okay, @elfmz if you do not sure this is acceptible to master, can we implement this as plugin? To do so, plugin API should be modified to allow plugins to intercept FileList::FindPartName, I guess

elfmz:
There is XLat function here that (was) doing similar thing: https://github.com/elfmz/far2l/blob/master/far2l/src/locale/xlat.cpp
I think its better to fix/adopt it as its used in different places including editor and may be used to transcode layout in editor, somehow (i didnt find yet how - from 1st look seems its accessible only from macroses)

elfmz:
ive created branch 'xlat', can you create PR into it? so i will continue adopting before merging it into master

elfmz:
aha, i can do it myself:)

PR #1221, opened 2022-01-30
minimal fix for ctrl+non_latin_keys on latest wx3.0

### LINKED ISSUE #1211: Ctrl/Alt+RussianLetter not working in wx3.0 with some keys
Alt+RussianLetter quick search is not working in wx3.0 with some keys. Whose are `б`, `ю`, `ж`, `э`, `х`, `ъ` at least.

Same problem with `Ctrl`+`ю` — marco recording do not starts in Russian keyboard layout. `Ctrl` hotkeys with other letter keys listed above does not work also.

unxed:
1. Implemented minimal fix for `Ctrl`+`б` `ю` `ж` `э` `х` `ъ` and possibly other non latin letters
2. Removed actually unneeded OnChar() call

Finally fixes #1211

### DIFF
--- a/WinPort/src/Backend/WX/wxMain.cpp
+++ b/WinPort/src/Backend/WX/wxMain.cpp
@@ -1063,6 +1063,25 @@ void WinPortPanel::OnKeyDown( wxKeyEvent& event )
 	if (alt_nonlatin_workaround) {
 		ir.Event.KeyEvent.wVirtualKeyCode = VK_UNASSIGNED;
 	}
+
+	if (event.ControlDown() && !event.GetKeyCode()) {
+		switch (event.GetRawKeyFlags()) {
+			// Hardware key codes are defined in /usr/share/X11/xkb/keycodes/xfree86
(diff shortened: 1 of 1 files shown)


unxed:
Possibly the same code should be in KeyUp(), but it works just as is and I wanted true minimal solution

unxed:
@elfmz may be we can make this feature optional, disabled by default, with enabling by flag file in config — for those who need it?

unxed:
@elfmz made this feature disabled by default, it is enabled by flag file in config (just as with "nomenu"). Now it is almost impossible for it to hurt anybody.

unxed:
Instead of hard coding x11 raw key codes to virtual key codes translations, we can use x11 to do it:
XKeycodeToKeysym(display, event.GetRawKeyFlags(), 0)) will give us X11 keysyms that can be mapped to virtual key codes without using hardcoded numbers.

unxed:
Closing in favor of #1569. It fixes the same problem better (and without hardcoded key code values).

Issue #1233, opened 2022-02-01
far2l silently erases characters in files in 1251 code page in editor if opened as utf8


unxed:
1. Unpack attached archive, open demo.php as UTF-8 in editor (it is in 1251 actually, that's important)
2. Do some changes
3. Save
4. Not only all chars with codes > 127 are erased (SILENTLY!), but also all string ends starting from first such character are silently erased also. What a mess! Even no notification of possible data loss :-/

It's very easy to miss that file is opened in wrong charset — especially if non-latin chars are not on the first screen. Any little change like commenting a line or two of sources — voila, lots of data are lost forever.

[demo.tar.gz](https://github.com/elfmz/far2l/files/7975403/demo.tar.gz)

unxed:
"Automatic detection" as default code page for editor may possibly solved this in most cases, but it does not work.

unxed:
@zerochka is also affected by this problem

akruphi:
About this problem I wrote as item 2 in https://github.com/elfmz/far2l/issues/1179

Now for apply change codepage in Editor need change UTF8->CP1251 then change from CP1251 to any other and return to CP1251.

unxed:
Looks like it is two closely related problems. My sample is rather small and it is impossible that problem is caused by slow file processing.

akruphi:
Additionally: when change codepage in Editor only by F8/Shift+F8 sometimes colorer incorrect highlight text and only setcodepage (by F8/Shift+F8), then exit and clearly re-open file by F4 help to correct codepage and colorer both.

unxed:
Reproduces on iconv branch also

unxed:
My small investigation shows that Far3 «converts» non-supported in UTF8 chars with codes from 128 to 255 to unicode 56448..56575 — this diapason is unused — and converts back on UTF8-to-ANSI/OEM conversion.

Can we behave just the same?

unxed:
Further investigation showed that far2l does not "see" the loss of characters during conversion if all lines in the file begin with characters with codes less than 128.

unxed:
Wrote an update to UTF-8 parsing code that mimics Far3 behavior on processing incorrect (for UTF-8) input. Now you can open ANSI/OEM source code as UTF-8, make any modifications to latin parts of it, save — and, WOW, no data loss any more! All ANSI/OEM chars (comments, string constants, etc) will be preserved. See PR #1237

unxed:
[...]

PR #1239, opened 2022-02-03
fallback to ANSI code page automatically if opening file in editor as UTF-8 fails

unxed:
yet another approach to minimize #1233's discomforts

why ANSI? ok, it can be made configurable. at least, it is 1-byte encoding which is guaranteed not to make any harm to users data


unxed:
Non needed any more after introducing bad utf8 sequences escaping in editor, see https://github.com/elfmz/far2l/commit/058ab0394bc50ee585da1f6029d867793fb00c53 and further commits

Issue #1222, opened 2022-01-30
[feature request] expand tabs option based on path


unxed:
putty's sources use spaces. far2l uses tabs. Continuously switching between modes produces mixed files, which is mess. It would be great to have an option to configure "expand tabs" editor option differently for different paths.

elfmz:
or may be even autodetect..

unxed:
default editor's charset by path is also appreciated

unxed:
Please implement if possible. Still hurts.

elfmz:
See branch edit-live-expandtab
There was imlemented changing tabs/spaces by F5 in editor and saving tabs expansion per-file in history.
Planned improvement is Shift+F5 to allow per-file spaces count modification on the fly and may be increasing history elements count from 512 to 4096 or so.. Later will cause old histories to be lost BTW.

unxed:
Great! Will look within a half of an hour. Maybe it is possible to detect indentation style for the first time opened files also?

unxed:
Looks working!

unxed:
With this simple addition I've got just that I wanted :)

```
diff --git a/far2l/src/fileedit.cpp b/far2l/src/fileedit.cpp
--- a/far2l/src/fileedit.cpp
+++ b/far2l/src/fileedit.cpp
@@ -1576,6 +1576,15 @@ int FileEditor::LoadFile(const wchar_t *Name,int &UserBreak)
 			LastLineCR=1;
 		}
 
+        // simple indentation style auto detection
+        if (Str[0] && (Str[0] == '\t')) {
+            m_editor->SetConvertTabs(0);
+        }
+        if (Str[0] && (Str[0] == ' ') && Str[1] && (Str[1] == ' ')) {
+            m_editor->SetConvertTabs(1);
+        }
+
+
 		if (!m_editor->InsertString(Str, StrLength))
 		{
 			EditFile.Close();
```

akruphi:
> See branch edit-live-expandtab There was imlemented changing tabs/spaces by F5 in editor and saving tabs expansion per-file in history. Planned improvement is Shift+F5 to allow per-file spaces count modification on the fly and may be increasing history elements count from 512 to 4096 or so.. Later will cause old histories to be lost BTW.

Extra fine! Very useful features! But it required **clear simple info** about current mode. In top status line in Editor some info will be useful:
- number of current tab spaces
- TAB as \t or expand to spaces
- sign about on/off tab autodetection

For example may be it write as:
[...]

PR #1268, opened 2022-02-19
better charset detection

unxed:
In Ubuntu 22.04 uchardet returns IBM852 on some source code files with pseudographic chars in comments. While not being perfect (it should be CP437), this guess is far better then WINDOWS-1252 (as in Ubuntu 20.04 on the same files) or UTF-8 (as if far2l do not understand uchardet return string).

This PR fixes far2l's understanding of uchardet return values like this.

### DIFF
--- a/far2l/src/locale/DetectCodepage.cpp
+++ b/far2l/src/locale/DetectCodepage.cpp
@@ -33,6 +33,9 @@ static int TranslateUDCharset(const char *cs)
 		return atoi(cs + 2);
 	}
 
+    if (strncasecmp(cs, "IBM", 3) == 0 && IsDecimalNumber(cs + 3)) {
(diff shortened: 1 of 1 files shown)


far2l/src/locale/DetectCodepage.cpp
```diff
@@ -33,6 +33,9 @@ static int TranslateUDCharset(const char *cs)
 		return atoi(cs + 2);
 	}
 
+    if (strncasecmp(cs, "IBM", 3) == 0 && IsDecimalNumber(cs + 3)) {
```
elfmz, line 36: pls use TABs

unxed: done!

#1222 😢

PR #1274, opened 2022-02-20
hackfix double escaping then running far2l inside far2l, fixes #1266

### LINKED ISSUE #1266: rendering artifacts then running tty far2l inside wx far2l
1. open far2l wx
2. run `far2l --tty` inside
3. open attched file in editor as UTF-8
[demo.txt](https://github.com/elfmz/far2l/files/8101882/demo.txt)
4. close it
5. you will see garbage in panel:
image

unxed:
Unfortunately, I lack understanding of far2l's internals to solve the problem perfectly. Namely, I do not understand very well at what points escape and de-escaping should be performed. However, this hack removes the observed problems, and does not seem to create new ones. See #1266 for details


elfmz:
Map be escaping is bad thing here at all and char mapping like used in far3 to handle bad UTF8 is more suitable here as it will not break layout due to some app outputs occasional wrong characters. Try vt-map-bad-utf branch

unxed:
> Try vt-map-bad-utf branch

#1266 not reproduced on that branch, great!

tried to print some invalid chars in terminal, looks like this:

```
/home/unxed/far2l$ echo проверка | iconv  -f=utf8 -t=cp866                                        ↑
஢ઠ
---------------------------------------------------------------------------------------------------
/home/unxed/far2l$ echo проверка | iconv  -f=utf8 -t=cp1251                                       ↑

```

 #1275 is still present though

unxed:
The theoretical problem with this approach is that if we have a file on disk with a name that is not a valid UTF-8 sequence, we will be able to see it in `ls`, but we will not be able to rename it with `mv`. So you probably need to add reverse translation for the command line. However, reverse translation has its drawbacks and can be harmful if the user himself uses the codes used for translation in command line for some reason.

However, this is a very far-fetched scenario, but the double escaping issue is very real and very likely to happen in everyday use.

elfmz:
for file names escaping still used, so it will work fine

elfmz:
ah in terminal, yes in terminal there will be problem, but in panels should be fine still)

PR #1294, opened 2022-03-16
Changed incsearch license to GPLv2

rampitec:


rampitec:
JFYI, I don't really care v2 or v3, but I have noticed v3 is an obstacle for Debian packaging.

Issue #1295, opened 2022-03-17
mounts sh freezes for network mounts


sxored:
have to wait from 30 sec to 2 minutes for cifs or sshfs mounts when remote host is unreachable

possible solution is to use mount command output for some info instead of df and after that use df command with predefined timeout for full info

i will submit pull request after some checks in different OS-es

sxored:
ready for new PR but what about PR #1289 ?

some changes are common to both PR and some ...

make new or continue in same branch ?

sxored:
tested new mounts.sh script with all changes in these environments

debian 10.11 , debian 11.2 , freebsd 12.1 and openwrt 19.x busybox shell 1.30.1

debian 10.11 @ after 3rd run

```
$ time sh ~/.config/far2l/mounts.sh
/	1.6G 9.6G 83%	    ext4
/usr/lib/modules	1.6G 9.6G 83%	 overlay
/run	2.4G 2.4G  1%	   tmpfs
-	home dirs
/home/user/Documents	=  1.1G	~/Documents
/home/user/Downloads	=  288M	~/Downloads
/home/user/Pictures	=   11M	~/Pictures
/home/user/Public	          	~/Public
-	home work
/home/user/projects	=  1.9G	~/projects
/home/user/various	=  -	~/various
-	home hidden
/home/user/.config	=  416M	~/.config
/home/user/.local/share	=   78M	~/.local/share
/home/user/.cache	=  237M	~/.cache
-	system
/tmp	=   68K	/tmp
/var/tmp	=   16K	/var/tmp
-	mnt
/	 /dev/sda	/
/home	 /dev/sdb	/home
/mnt/backup	  +	/mnt/backup
/mnt/custom	  +	/mnt/custom
/mnt/floppy	  +	/mnt/floppy
/mnt/media	  -	/mnt/media
/mnt/cdrom	    @cdrom	/mnt/cdrom
/mnt/remote1	  +	/mnt/remote1
/mnt/remote2	  +	/mnt/remote2

real	0m0.221s
user	0m0.097s
sys	0m0.145s

```
debian 11.2 @ 1st run after boot

```
$ time ~/.config/far2l/mounts.sh
/	2.7G 9.6G 70%	    ext4
/usr/lib/modules	2.7G 9.6G 70%	 overlay
/run	785M 786M  1%	   tmpfs
/mnt/backup	6.2G 176G 97%	    ext4
/mnt/custom	4.9G 147G 97%	    ext4
-	home dirs
[...]

PR #1306, opened 2022-04-02
fixed mounts sh freezing for network mounts

sxored:
fix for #1295

### DIFF
--- a/far2l/bootstrap/mounts.sh
+++ b/far2l/bootstrap/mounts.sh
@@ -143,18 +143,20 @@ else
 			avail_fraction = avail < 10 ? 1 : 0;
 			total_fraction = total < 10 ? 1 : 0;
 
+			separator_symbol = "/"
+
 			if ("'$awk_busybox'" == "true") {
 
 				avail2 = substr((avail_ident avail ".0"), 1, 3);
 				total2 = substr((total_ident total ".0"), 1, 3);
 					
-				print_format = "%s\t%s%s %s%s %s\t%s\n";
-				printf print_format, path, avail2, avail_units, total2, total_units, $'$DF_USE', $'$DF_NAME';
+				print_format = "%s\t%s%s%s%s%s\t%s\n";
+				printf print_format, path, avail2, avail_units, separator_symbol, total2, total_units, $'$DF_NAME';
 
 			} else {
 
-				print_format = "%s\t%s%.*f%s %s%.*f%s %3s\t%8s\n";
-				printf print_format, path, avail_ident, avail_fraction, avail, avail_units, total_ident, total_fraction, total, total_units, $'$DF_USE', $'$DF_NAME';
+				print_format = "%s\t%s%.*f%s%s%s%.*f%s\t%8s\n";
+				printf print_format, path, avail_ident, avail_fraction, avail, avail_units, separator_symbol, total_ident, total_fraction, total, total_units, $'$DF_NAME';
 			}
 		}
 	}'
--- a/far2l/bootstrap/mounts.sh
+++ b/far2l/bootstrap/mounts.sh
@@ -42,22 +42,78 @@ if [ "$1" = 'umount' ]; then
 ##########################################################
 else
 	sysname="$(uname)"
+	timeout_coreutils="false"
+	column_bsdmainutils="false"
+	awk_busybox="false"
+	if [ "$sysname" = "Linux" ]; then
+		timeout_coreutils="$(sh -c 'timeout --version > /dev/null 2>&1 && printf true || printf false')"
+		column_bsdmainutils="$(sh -c 'column -h > /dev/null 2>&1 && printf true || printf false')"
+		awk_busybox="$([ .$(awk 2>&1 | head -n 1 | cut -d' ' -f1) = .BusyBox ] > /dev/null 2>&1 && printf true || printf false)"
+	fi
 	if [ "$sysname" = "Linux" ] || [ "$sysname" = "FreeBSD" ]; then
 		DF_ARGS='-T'
 		DF_AVAIL=5
 		DF_TOTAL=3
 		DF_NAME=2
 		DF_DIVBY=1
+		DF_USE=6
+		MNT_FS=1
+		MNT_PATH=3
+		FT_IDX=1
+		FT_FS=3
+		FT_TYPE=4
+		FT_PATH=5
+		TIMEOUT=1
+		MNT_TYPE=0
+		MNT_FILTER='cat'
+		FT_HEADER='Filesystem Type 1K-blocks Used Avail Use%% Mountpoint'
+		FMT_COLUMN='cat'
+		if [ "$sysname" = "Linux" ]; then
+			MNT_TYPE=5
+			MNT_FILTER='grep -v -e " /proc\(/[\/a-z_]\+\)* " -e " /sys\(/[a-z_,]\+\)* " -e " /dev/\(pts\|hugepages\|mqueue\) " -e " /run/user/\([0-1]\+\)/doc " -e "/home/\([a-z0-9\._]\+\)/.cache/doc"'
+			FT_HEADER='Filesystem Type 1K-blocks Used Available Use%% Mounted_on'
+			if [ "$timeout_coreutils" = "true" ]; then
+				FMT_COLUMN='column -t'
+			fi
+		fi
+		if [ "$sysname" = "FreeBSD" ]; then
+			MNT_FILTER='grep -v -e " \(/[a-z]\+\)*/dev/foo "'
+			FT_HEADER='Filesystem Type 1K-blocks Used Avail Capacity Mounted_on'
+		fi
+
+		# 1} printf $FT_HEADER and mount output
+		# 2) grep - use $MNT_FILTER
+		# 3) awk - output almost like df
+		# 4) sed - prepare for uniq to eliminate duplicated block device names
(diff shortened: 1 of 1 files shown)


elfmz:
Sorry but i'd like simple possible solution, please check recent commit if it resolves original issue

sxored:
also this simpler approach does not work as expected when there are more than one network mounts to different hosts

it does not show any network mount in list when only one of them is not reachable

while mount command work flawlessly in almost all cases and OS-es

that's why i extended code with checks and gathering info from df command for every mounted device one by one ...

sxored:
please see this

https://stackoverflow.com/questions/601543/command-line-command-to-auto-kill-a-command-after-a-certain-amount-of-time

sxored:
previous link is from this answer

https://stackoverflow.com/questions/37981492/how-to-timeout-a-group-of-commands-in-bash/61888916#61888916

elfmz:
Yes, df -l abscent on busybox, but it works on MacOS, while your script fallbacks to old behavior there.
And still need to think how to simplify it. This shell script appeared cuz it was too complicated to write in C++ due to there is no universal way to get mount points across all OSes, but now its also looks complicated..So may be consider switch to C++ ) 
Few obvious things:
1 FT_HEADER not needed (at least it was not needed for code that parses output of df)
2 indent awk scripts
...and remove those underscores in disk space numbers.

sxored:
one can use uniq and sort commands to remove "duplicates" ( they are not really duplicates, only by device name ) in mount command output to imitate df output ...

but there are different implementations of this commands in various environments ( gnu/nongnu/bsd/etc )

fortunately awk, grep and sed have more or less similar implementation almost everywhere )

and we can use them with help of "uniq -w NN" for this task

```
$ mount | grep -v -e " /proc\(/[\/a-z_]\+\)* " -e " /sys\(/[a-z_,]\+\)* " -e " /dev/\(pts\|hugepages\|mqueue\) " -e " /run/user/\([0-1]\+\)/doc " | awk '{ print $1 " $ " $1 " " $5 " " $3 " " $NF }' | grep -n -e "" | sed -e "s_^\([1-9]:\)_0\1_" | sed -e "s_^\([0-9][0-9]:\)_0\1_" | sed -e "s_^\([0-9]\+\):\(\/dev\/[^\ ]\+\)_000:~\2_" | column -t

000:~/dev/vda1           $  /dev/vda1            ext4      /                 (rw,noatime,discard)
002:none                 $  none                  overlay   /usr/lib/modules  (rw,relatime,lowerdir=/tmp/modules,upperdir=/sysroot/lib/modules,workdir=/sysroot/lib/.modules_work)
003:devtmpfs             $  devtmpfs              devtmpfs  /dev              (rw,nosuid,noexec,size=2000228k,nr_inodes=500057,mode=755,inode64)
004:tmpfs                $  tmpfs                 tmpfs     /dev/shm          (rw,size=1048576k,inode64)
005:tmpfs                $  tmpfs                 tmpfs     /run              (rw,nosuid,nodev,size=804928k,nr_inodes=819200,mode=755,inode64)
006:tmpfs                $  tmpfs                 tmpfs     /run/lock         (rw,nosuid,nodev,noexec,relatime,size=5120k,inode64)
000:~/dev/vdb1           $  /dev/vdb1             ext4      /home             (rw,relatime,discard)
000:~/dev/vdb1           $  /dev/vdb1             ext4      /usr/local        (rw,relatime,discard)
000:~/dev/vdb1           $  /dev/vdb1             ext4      /var/spool/cron   (rw,relatime,discard)
011:tmpfs                $  tmpfs                 tmpfs     /run/user/1000    (rw,nosuid,nodev,relatime,size=402460k,nr_inodes=100615,mode=700,uid=1000,gid=1000,inode64)
000:~/dev/vdi            $  /dev/vdc1             fuseblk   /mnt/media        (rw,nosuid,nodev,noexec,relatime,user_id=0,group_id=0,default_permissions,allow_other,blksize=4096,user)
000:~/dev/vdk            $  /dev/vdd1             ext4      /mnt/backup       (ro,nosuid,nodev,noexec,relatime,norecovery,user)
000:~/dev/vdj            $  /dev/vde1             ext4      /mnt/custom       (rw,nosuid,nodev,noexec,relatime,user)
015://a.b.c.d/share      $  //a.b.c.d/share       cifs      /mnt/remote1      (rw,nosuid,nodev,noexec,relatime,vers=2.1,cache=strict,...)
016://a.b.c.d/share      $  //a.b.c.d/share       cifs      /mnt/remote2      (rw,nosuid,nodev,noexec,relatime,vers=2.1,cache=strict,...)
```

"column -t" is for human eyes only and is not necessary

i do not forgot about "duplicated" network mounts

in far2l one can unmount multiple network mounts separately

```
$ mount | grep -v -e " /proc\(/[\/a-z_]\+\)* " -e " /sys\(/[a-z_,]\+\)* " -e " /dev/\(pts\|hugepages\|mqueue\) " -e " /run/user/\([0-1]\+\)/doc " | awk '{ print $1 " $ " $1 " " $5 " " $3 " " $NF }' | grep -n -e "" | sed -e "s_^\([1-9]:\)_0\1_" | sed -e "s_^\([0-9][0-9]:\)_0\1_" | sed -e "s_^\([0-9]\+\):\(\/dev\/[^\ ]\+\)_000:~\2_" | uniq -w 16 | awk '{ print $3 " " $4 " [size] [used] [avail] [use] " $5 " " $NF }' | column -t

/dev/vda1            ext4      /                 (rw,noatime,discard)
none                  overlay   /usr/lib/modules  (rw,relatime,lowerdir=/tmp/modules,upperdir=/sysroot/lib/modules,workdir=/sysroot/lib/.modules_work)
devtmpfs              devtmpfs  /dev              (rw,nosuid,noexec,size=2000228k,nr_inodes=500057,mode=755,inode64)
tmpfs                 tmpfs     /dev/shm          (rw,size=1048576k,inode64)
tmpfs                 tmpfs     /run              (rw,nosuid,nodev,size=804928k,nr_inodes=819200,mode=755,inode64)
tmpfs                 tmpfs     /run/lock         (rw,nosuid,nodev,noexec,relatime,size=5120k,inode64)
/dev/vdb1             ext4      /home             (rw,relatime,discard)
tmpfs                 tmpfs     /run/user/1000    (rw,nosuid,nodev,relatime,size=402460k,nr_inodes=100615,mode=700,uid=1000,gid=1000,inode64)
/dev/vdc1             fuseblk   /mnt/media        (rw,nosuid,nodev,noexec,relatime,user_id=0,group_id=0,default_permissions,allow_other,blksize=4096,user)
/dev/vdd1             ext4      /mnt/backup       (ro,nosuid,nodev,noexec,relatime,norecovery,user)
/dev/vde1             ext4      /mnt/custom       (rw,nosuid,nodev,noexec,relatime,user)
//a.b.c.d/share       cifs      /mnt/remote1      (rw,nosuid,nodev,noexec,relatime,vers=2.1,cache=strict,...)
//a.b.c.d/share       cifs      /mnt/remote2      (rw,nosuid,nodev,noexec,relatime,vers=2.1,cache=strict,...)
```

sxored:
> ... due to there is no universal way to get mount points across all OSes, but now its also looks complicated..So may be consider switch to C++ )

why not ?

mount and df command implementations are almost all opensource these days )

> Few obvious things:
> 1 FT_HEADER not needed (at least it was not needed for code that parses output of df)

for human eyes only

> 2 indent awk scripts

i will reformat code

> ...and remove those underscores in disk space numbers.

it is necessary to indent the first number by some "non-space" symbol, because far2l removes all the first "space" symbols ...

sxored:
> ... And still need to think how to simplify it. This shell script appeared cuz it was too complicated to write in C++ ...

Everything should be made as simple as possible, but not simpler. ( Albert Einstein? Louis Zukofsky? Roger Sessions? William of Ockham? Anonymous? )

(https://quoteinvestigator.com/2011/05/13/einstein-simple/#comment-4418)

> Eric says:
> [November 19, 2012 at 10:32 pm]

> The Einstein quote does NOT mean “make it simple enough to understand,” as Dodd suggested above. Einstein spoke and wrote more than once about simplification. More to the point, he was decidedly AGAINST simplifying to the point that laypersons might understand ...

sxored:
> Yes, df -l abscent on busybox, but it works on MacOS, while your script fallbacks to old behavior there.

This is because I have not yet had the opportunity to test the implementation of these programs ( mount, df, sed, grep, sort, uniq, awk, timeout ) on macOS

sxored:
interensting reading

https://gobolinux.org/doc/articles/clueless.html

> Fortunately, like with the live CD, we have nowadays a technological advancement that serves as a real solution to the problem: union mounts, also known as overlay filesystems. The idea is that you can mount several partitions in the same directory. This way, the semantics of /Programs as “the collection of all programs available in the system” is retained, independently of the physical location of the actual data. File systems are all about abstraction (we don't refer to files based on their track, sector and cylinder address), this progresses a step further. Overlay filesystems are very flexible: the sysadmin, for example, can overlay site-specific settings for an application on top of the defaults exported over NFS. Unfortunately, it is not in widespread use, for reasons beyond my understanding. The Plan 9 operating system has it as one of its basic filesystem operations: the bind command (in Plan 9, for example, you don't need a $PATH variable, because all directories containing executables are “bound” in a single directory). There is an implementation of an overlay filesystem for Linux: ovlfs.

sxored:
df -l may take a long time to process the request also in case of faulty local storage

I know it's rare, but not for system administrators using far2l to save some files

PR #1344, opened 2022-07-23
update mounts.sh regex - exclude tmp mounts

sxored:
resolves issue #1315

P.S. what about user regexes in mounts.sh ?


sxored:
```
 -e " /tmp/\.\([a-z0-9\._]\+\)_\([a-zA-Z0-9\._]\+\) "
```

regex excludes these example mounted directories from final output of mounts.sh

```
/tmp/.veracrypt_aux_mnt1
```

and

```
/tmp/.mount_firefoMAvMq7
```

PR #1368, opened 2022-10-04
Bunch of minor improvement

klirichek:
* Changed build instruction in Readme.md - to use uniform cmake calls for build/packaging
* Invoke 'cmake -E' for commands like 'rm', 'ln'
* Use explicit WORKING_DIRECTORY instead of 'cd ... && XXX'
* direct creation of symlinks for askpass/sudoapp for packaging, and remove prerm/postinst stuff because of it as redundant
* add 'copyright' file into debian package

As a side note - rpm may be as well be created without any additional stuff after build with 'cpack -G RPM', however it may need some adjustments in meta-information. M.b. even better to provide general CPACK_XXX variables, and then both deb and rpm can use these values as defaults, instead of providing a copy for each generator.


elfmz:
git should not be required to build far2l. It just provides precise version info. No git - no that info, but dont fail build plz

PR #1440, opened 2022-12-03
more busybox detection tweaks for compatibility with recent dd-wrt

unxed:


NetRocks/src/Protocol/SSH/ProtocolSCP.cpp
```diff
@@ -657,7 +663,7 @@ ProtocolSCP::ProtocolSCP(const std::string &host, unsigned int port,
 		}
 		if (std::find(words.begin(), words.end(), _quirks.rm_dir) == words.end()) {
 			fprintf(stderr, "ProtocolSCP: '%s' unsupported\n", _quirks.rm_dir);
-			_quirks.rm_dir = "rm -f -d";
+			_quirks.rm_dir = "rm -f -r"; // using -r as -d not supported on some devices
```
elfmz, line 666: but its unsafe to use -r - may occasionally (due to bug or misuse) delete whole tree

unxed: Reverted -d -> -r change. No ideas for now how to fix deletion on routers that do not support -d.

Btw, won't -d leave "orphan" file nodes in filesystem?

elfmz: what it have at all? may be rmdir? or rm without any flags removes directories?

unxed: > what it have at all? may be rmdir? or rm without any flags removes directories?

rmdir works (for empty dir only). rm without flags does not.

unxed: Added some fixes.
1) Now we do not try to analyze applets list if `busybox` command returned none
2) Now we detect unlink presence even if got no applets list from busybox

Now on fresh ddwrt not only busybox is detected fine, but also files/dirs removal works ok.

PR #1445, opened 2022-12-04
alternate default tar command lines for embedded builds

unxed:


elfmz:
better to use dedicated macro like __TAR_LIMITED_ARGS__ and pass it externally via CMake. I suspect its a tar that comes from busybox that can be used with glibc too, so need to enable this arg from BR confif if busybox is enabled

unxed:
done

PR #1511, opened 2023-02-22
allow WM to select window position, and restore possibility to set window size in chars

akozlins:
- first commit uses wxDefaultPosition for WinState::pos
  to allow window manger to select startup position if winstate file contains -1 for x/y positions
  (always using fixed position is not always desirable)

- second commit restores functionality to set windows size in characters if consolesize file is present
  (i use same config for size of windows in chars across different machines,
  and size in pixel is not good enough, also because it is difficult to resize window to get rid of black padding
  if size is not multiple of chars) 

possible modification (if requested):
- move consolesize values to winstate (but consolesize was already present in past versions)

### DIFF
--- a/WinPort/src/Backend/WX/wxMain.cpp
+++ b/WinPort/src/Backend/WX/wxMain.cpp
@@ -167,31 +167,25 @@ WinState::WinState()
 	fullscreen = (i & 4) != 0;
 
 	getline(is, str);
-	i = atoi(str.c_str());
-	if (i >= 100) {
-		size.SetWidth(i);
-	}
+	int width = atoi(str.c_str());
+	if (width >= 100)
+		size.SetWidth(width);
 
 	getline(is, str);
-	i = atoi(str.c_str());
-	if (i >= 100) {
-		size.SetHeight(i);
+	int height = atoi(str.c_str());
+	if (height >= 100)
+		size.SetHeight(height);
+
+	// interpret negative widht/height values as size in chars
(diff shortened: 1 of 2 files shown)


WinPort/src/Backend/WX/wxMain.cpp
```diff
@@ -182,6 +182,16 @@ WinState::WinState()
 	pos.x = atoi(str.c_str());
 	getline(is, str);
 	pos.y = atoi(str.c_str());
+
+	std::ifstream isconsize;
+	isconsize.open(InMyConfig("consolesize").c_str());
```
elfmz, line 187: consolesize is an old file where size was saved/restored. Using it here like this will lock all user who had old versions of far2l to same startup window size. So plz use different name, like "winsize" or "fixedsize" so it wil not break those who dont know about such file at all.

akozlins: as the position and size in winstate are saved when user does F9->Options->Save_Setup, how about using the consolesize values when x/y sizes in winstate are negative?
this will also solve the 'issue' that existing consolesize for 'old' user will still work as it was before removal if one manually sets values in winstate (can also be documented)

elfmz: I'd suggest in such case to not use consolesize and keep everything in single winstate file, so if size in winstate is negative - then treat it as characters but not pixels and don't change it on window size change.

PR #1527, opened 2023-03-07
Added .editorconfig files for consistent code style

unxed:


unxed:
Files/folders with mixed indentation that possibly needs to be reformatted using consistent code style:
```
WinCompat.h
inside/CMakeLists.txt
7z.cpp
elfinfo.h
editorcomp/src
several modules from cmake/modules
several files from calc plugin sources
bootstrap/CMakeLists.txt
```

PR #1547, opened 2023-03-12
Fix mixed indentation

unxed:


WinPort/WinPort.h
```diff
@@ -21,7 +21,7 @@ extern "C" {
 	void WinPortHelp();
 	const wchar_t *WinPortBackend();
 
-	///console API
```
elfmz, line 24: actually triple / is not a mistake - it produces special coloring:
Screenshot_20230312_182100

unxed: fixed that

PR #1548, opened 2023-03-12
Fixed mixed indentation

unxed:


elfmz:
May be no need to change commenting kinds? As its prone to manual errors, i don't see any problems with having both kinds.

PR #1555, opened 2023-03-14
Code readability

unxed:


elfmz:
I don't like splitting to many too short lines in same degree as too long lines. Suggest to setup few rules:
- Sane characters count per line is about 90. This is a flexible 'attractor' value, but not hardcoded limit.
- In case having multiple arguments, some of which are semantically tied more then others - prefer to keep them together at same line. e.g. `const char *buffer, size_t buffer_length`

elfmz:
Also regarding readability - biggest issue with code is spacing, again, need fisrts to setup some rules before making changes. I'd like not to verbalize that but use simple examples:
int foobar(char *arg1, size_t arg2)
{
  if (arg1 && arg2 && strcmp(arg1, arg2) == 0) {
    std::cout << "Both args have value" << arg1 << std::endl;
  }
}

unxed:
Having rules is for sure good idea!

Btw, yours there are some rather complex cases. How do you prefer formatting code like this?
```
	if (Key!=KEY_NONE && Key!=KEY_IDLE && Key!=KEY_SHIFTINS && Key!=KEY_SHIFTNUMPAD0 && Key!=KEY_CTRLINS &&
		((unsigned int)Key<KEY_F1 || (unsigned int)Key>KEY_F12) && Key!=KEY_ALT && Key!=KEY_SHIFT &&
		Key!=KEY_CTRL && Key!=KEY_RALT && Key!=KEY_RCTRL &&
		(Key<KEY_ALT_BASE || Key > KEY_ALT_BASE+0xFFFF) && // ???? 256 ???
		!(((unsigned int)Key>=KEY_MACRO_BASE && (unsigned int)Key<=KEY_MACRO_ENDBASE) || ((unsigned int)Key>=KEY_OP_BASE && (unsigned int)Key <=KEY_OP_ENDBASE)) && Key!=KEY_CTRLQ)
	{
		//...
	}
```
Here we can not find any line splitting logic as all conditions are semantically tied. I used the rule "two conditions per line" to avoid too short lines:
```
	if (
		Key!=KEY_NONE && Key!=KEY_IDLE &&
		Key!=KEY_SHIFTINS && Key!=KEY_SHIFTNUMPAD0 &&
		Key!=KEY_CTRLINS && ((unsigned int)Key<KEY_F1 || (unsigned int)Key>KEY_F12) &&
		Key!=KEY_ALT && Key!=KEY_SHIFT &&
		Key!=KEY_CTRL && Key!=KEY_RALT &&
		Key!=KEY_RCTRL && (Key<KEY_ALT_BASE || Key > KEY_ALT_BASE+0xFFFF) && // ???? 256 ???
		!(
			((unsigned int)Key>=KEY_MACRO_BASE && (unsigned int)Key<=KEY_MACRO_ENDBASE) ||
			((unsigned int)Key>=KEY_OP_BASE && (unsigned int)Key <=KEY_OP_ENDBASE)
		) && Key!=KEY_CTRLQ
	) {
		//...
	}
```

In this example the longest line has 92 chars width, which is very close to 90 that you suggested. Is such fomatting ok?

elfmz:
`Key!=KEY_NONE` -> `Key != KEY_NONE` and `Key > KEY_ALT_BASE+0xFFFF` -> `Key > KEY_ALT_BASE + 0xFFFF`
This will make it even longer but still its ok. 90 - is not hard limit, but soft watermark.

elfmz:
Also, it makes sense to do such spaces adjusting by clang-format but not manually..

unxed:
Had some experiments with clang-format. Could not find configuration options to completely prevent it from introducing mixture of tabs and spaces in the beginning of lines — the thing I was trying to avoid. Tried this config:
```
---
IndentWidth: 4
TabWidth: 4
UseTab: Always
AlignAfterOpenBracket: DontAlign
ColumnLimit: 120
AlignOperands: DontAlign
```
Still things like
```
Edit::Edit(ScreenObject *pOwner, Callback *aCallback, bool bAllocateData)
	: m_next(nullptr), m_prev(nullptr),
	  Str(bAllocateData ? reinterpret_cast<wchar_t *>(malloc(sizeof(wchar_t))) : nullptr), StrSize(0), MaxLength(-1),
	  Mask(nullptr), LeftPos(0), CurPos(0), PrevCurPos(0), MSelStart(-1), SelStart(-1), SelEnd(0), CursorSize(-1),
	  CursorPos(0) {
```
continue to happen.

unxed:
Found astyle parameters combination that gives almost that we want:
```
astyle -T -S -p -xg -H -U -xU filename.cpp
```

PR #1569, opened 2023-03-26
Fix zero virtural key code for Alt+non_latin_letter key combinations

### LINKED ISSUE #1562: fix Alt+letter hot keys in non-latin layout in turbo under far2l wx
There is a problem using Alt+letter hot keys then running [turbo](https://github.com/magiblot/turbo) under far2l wx: they do not work in non-latin keyboard layout. This is a complex problem and need solving on both far2l and turbo sides.

First of all, wx do not send key codes in such cases at all (event.GetKeyCode() is zero for Alt+non_latin_letter). So we should detect corresponding latin letter on far2l side:
```
wx2INPUT_RECORD::wx2INPUT_RECORD(BOOL KeyDown, const wxKeyEvent& event, const KeyTracker &key_tracker)
{
	auto key_code = event.GetKeyCode();

#ifdef __WXGTK__
	if (key_code == 0) {
	    Display *display;
	    display = XOpenDisplay(NULL);
	    // GetRawKeyFlags() gives us raw x11 hardware key code under wxGTK
	    char* keysym = XKeysymToString(XKeycodeToKeysym(display, event.GetRawKeyFlags(), 0));
	    if (strlen(keysym) == 1) {
	    	// char key
	    	key_code = toupper(*keysym);
	    }
	    XCloseDisplay(display);
	}
#endif
```

On the turbo side, we should use wVirtualKeyCode value to detect such hot keys instead of uChar.UnicodeChar value. @magiblot can we implement this?

unxed:
Fixes #1562 and #1211. Replaces #1221.


unxed:
FYI: as long as https://github.com/wxWidgets/wxWidgets/commit/2c0f6a2aa0c8d0650d1f856aa279ce7495925c17 is merged to wx master , we do not need all this `__HASX11__` stuff. At leasts, we can safely wrap it into "only use it with wx < 3.3.0" ifdefs to avoid x11 dependency on Wayland-only systems. Although the dependency is not mandatory and can simply be left as is.

PR #1573, opened 2023-03-29
Better key_code detection

unxed:
The previous attempt to detect key_code could behave incorrectly if the system has Latin keyboard layouts other than English. For example, on the German keyboard layout, the keys like `;` or `'` are placed differently from English one, also do `z` and `y` keys.

So I modified the code so that the detection of key_code will only work if the system has English keyboard layout installed, and only English keyboard layout would be used for detection.

Better approach would be to use xkbcommon lib (as it allows to do key mapping using keyboard layouts not enabled system wide; [example](https://github.com/wxWidgets/wxWidgets/issues/23379#issuecomment-1488563133)), but it is one more dependency so we probably do not want it.


elfmz:
I've added some caching to your workaround, plz check if it still works correctly.
Caching is necessary as there were users who are using far2l in remote X11 session and in such case all that X11 calls could introduce noticeable delays to handling keypresses, that is annoying.

unxed:
Still works as expected, thanks!

PR #1574, opened 2023-03-30
kb layout independent keystroke info for tty backend also. see #1562

unxed:
also re-enables compact input mode

### DIFF
--- a/WinPort/src/Backend/TTY/TTYX/TTYX.cpp
+++ b/WinPort/src/Backend/TTY/TTYX/TTYX.cpp
@@ -39,6 +39,7 @@ class TTYX
 	Atom _clipboard_atom;
 	Atom _xsel_data_atom;
 	int _display_fd;
+	int _eng_group_id = 0; // use default kb layout if English one is not found
 
 #ifdef TTYXI
 	const std::chrono::time_point<std::chrono::steady_clock> _never;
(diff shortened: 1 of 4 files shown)


WinPort/src/Backend/TTY/TTYX/TTYX.cpp
```diff
@@ -155,7 +155,26 @@ class TTYX
 			if ((cookie->evtype == XI_RawKeyPress || cookie->evtype == XI_RawKeyRelease) && cookie->data) {
 				const XIRawEvent *ev = (const XIRawEvent *)cookie->data;
 				//fprintf(stderr, "TTYXi: !!!!! %d %d\n", cookie->evtype == XI_RawKeyPress, ev->detail);
-				const KeySym ks = XkbKeycodeToKeysym(_display, ev->detail, 0 /*xkbState.group*/, 0 /*shift level*/);
+
+				XkbDescPtr xkb = XkbGetMap(_display, 0, XkbUseCoreKbd);
```
elfmz, line 159: its better to call XkbGetMap/XkbGetControls/XkbGetNames in constructor of  TTYX remembering all needed things and do here only XkbKeycodeToKeysym

unxed: If done this way, there may be errors if the user adds or removes keyboard layouts while far2l is already running.

elfmz: I think its a very rare situation and user will restart far2l before posting bug here)
But once somebody will post a bug, will then use https://tronche.com/gui/x/xlib/events/window-state-change/mapping.html

unxed: ok, fixed!

unxed: Is there something else left to be done?

unxed:
oh, there is a build problem in some configurations, probably fixed it in #1576

unxed:
sorry, one more build problem, see #1577

PR #1582, opened 2023-04-02
Win32 input mode: fixed some bugs found while testing on real Windows Terminal

unxed:
In particular,

1. input buffer can contain more than one esc seq
2. respect buffer size (l variable)
3. win32 input seq length can be more than 8 bytes
4. do not mess win32 input with mouse coords events

Original bug report:
https://t.me/far2l_ru/6587

See also:
https://github.com/microsoft/terminal/issues/15083

### DIFF
--- a/WinPort/src/Backend/TTY/TTYInputSequenceParser.cpp
+++ b/WinPort/src/Backend/TTY/TTYInputSequenceParser.cpp
@@ -311,9 +311,11 @@ void TTYInputSequenceParser::ParseAPC(const char *s, size_t l)
 
 size_t TTYInputSequenceParser::ParseEscapeSequence(const char *s, size_t l)
 {
-	if (s[0] == '[' && s[l-1] == '_') {
+    const char* underscore_pos = strchr(s, '_');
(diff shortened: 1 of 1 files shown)


elfmz:
There re still problems with such approach as it doesnt check sequence to be valid and can accept (and consome as whole) as winterm any multiple sequences that starts by [ and has _ inside.
Also i'm not sure about safety of using sscanf for case where integer parts can be empty or some missing at all.
So i've just rewrote that parsing to IMHO safer manner but cannot test if it still works..

unxed:
Looks like there is still a problem with
```
	return (l >= 8) ? TTY_PARSED_BADSEQUENCE : TTY_PARSED_WANTMORE;
```

We should increase 8 to 32 as win32-input-mode esc seqs are longer than 8 chars. This 8 byte limitation already caused bugs in my experiments.

unxed:
The man who suffered from problems in WinTerm tested it, https://github.com/elfmz/far2l/commit/b5f7ba323e61370b62caeda59b05970405fa7abd works just as well as the one I suggested. So closing, thanks!

PR #1583, opened 2023-04-02
latin key guess to work even if english kb layout not installed

unxed:
I understood how to do this with the xkbcommon API, but didn't want to add a new dependency. Luckily, ChatGPT 4 was able to tell me how to do the same on a pure x11 api. Genius tool :)

### DIFF
--- a/WinPort/src/Backend/WX/wxWinTranslations.cpp
+++ b/WinPort/src/Backend/WX/wxWinTranslations.cpp
@@ -421,47 +421,61 @@ static int X11KeyCodeLookupUncached(wxUint32 keyflags)
 {
 	int key_code = 0;
 	Display *display = XOpenDisplay(NULL);
-	XkbDescPtr xkb = XkbGetMap(display, 0, XkbUseCoreKbd);
+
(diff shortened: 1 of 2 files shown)


WinPort/src/Backend/WX/wxWinTranslations.cpp
```diff
@@ -421,47 +421,61 @@ static int X11KeyCodeLookupUncached(wxUint32 keyflags)
...
-		if (strstr(layout, "English")) {
-			// English kb layout found, let's use it for translations
-			group = i;
-			break;
-		}
-		XFree(layout);
-	}
+	KeySym ks;
+	unsigned int mods;
+	XkbTranslateKeyCode(xkb, keyflags, 0, &mods, &ks);
```
elfmz, line 454: missing error check: XkbTranslateKeyCode returns True if the translation resulted in a keysym, and False if it resulted in NoSymbol.

unxed: oh, thanks for noticing and fixing!

PR #1601, opened 2023-04-09
.editorconfig settings for some files that were missed previously

unxed:


elfmz:
Tabs in makefiles have special meaning, so don't do any restrictions for them: https://openjdk.org/groups/build/doc/code-conventions.html

elfmz:
https://retrocomputing.stackexchange.com/questions/20292/why-does-make-only-accept-tab-indentation

unxed:
thanks! removed makefiles part

Issue #1575, opened 2023-03-30
support "rich keyboard" protocols by iTerm2/kovidgoyal's kitty/Windows Terminal


unxed:
At the time far2l tty mode was firstly introduced, no terminal supported sending rich keypress information to console apps. Now it is possible in at least two terminals: iTerm2 on Mac and kovidgoyal's kitty on Mac/BSD/Linux.

Each one uses its own protocol for that (actually there is also third protocol for that, called "CSI u" or "libtickit protocol" and supported by some other terminals, but it does not suit far2l well as it is unable to tell equivalent latin letter for alt/ctrl+non_latin_letters key combinations, also no separate escape sequences for keyup/keydown events and some other limitations).

The first one is **iTerm2 "raw keyboard" protocol**:
https://gitlab.com/gnachman/iterm2/-/issues/7440#note_129307021

The second one is **kovidgoyal's kitty keyboard protocol**:
https://sw.kovidgoyal.net/kitty/keyboard-protocol/

If far2l's tty backend will support those portocols it will for sure improve UX of tty version users.

PS: Maybe iTerm2's protocol should not have priority in implementation as it looks like they are planning to adopt kovidgoyal kitty's one also sometime in the future:
https://gitlab.com/gnachman/iterm2/-/issues/10017

unxed:
kovidgoyal's kitty keyboard protocol is also discussed for inclusion in Windows Terminal. I appreciate voting for it instead of "CSI u" as it fits far2l better:
https://github.com/microsoft/terminal/issues/11509#issuecomment-1487289848

unxed:
Oh, Windows Terminal has it's own "rich keyboard" protocol, `_KEY_EVENT_RECORD`s are sent as escape sequences!

https://github.com/microsoft/terminal/blob/main/doc/specs/%234999%20-%20Improved%20keyboard%20handling%20in%20Conpty.md

unxed:
> Oh, Windows Terminal https://github.com/microsoft/terminal/pull/6309, fields from _KEY_EVENT_RECORD are sent inside special escape sequence!

Implemented it, see #1581

unxed:
while testing win32-input-mode introduced in https://github.com/elfmz/far2l/commit/edf9b7397828b6c6366081c95751f09d1c45b87d under Windows Terminal, noticed very strange Windows Terminal behavior: https://github.com/microsoft/terminal/issues/15083

unxed:
[...]

Issue #1101, opened 2021-10-10
Bring TUI version to functional parity with GUI version as close as possible


unxed:
Some users do not want to leave their preferred terminal apps. For whose making TUI version better will bring greater experience, spreading far2l usage wider. Most of this work is already done, but there are still some minor issues left.

1. Complex hotkeys, like Ctrl+Shift+arrows. Already supported in terminal. Single ESC press — also supported (an option in settings to enable it by default is appreciated).

2. Mouse. Also already works in terminal.

3. Some terminal apps intercept some hotkeys used by far2l. It may be disabled in their settings (instructions on doing this may be added to far2l help optionally).

4. X clipboard access. If far2l extensions are not available, we should fall back to xclip tool usage (AFAIK mc is already doing something like this). There is also `OSC 52` for clipboard access, but it looks like it will never be supported by terminals, so we may postpone its support for now:
https://gitlab.gnome.org/GNOME/vte/-/issues/2495
https://bugs.kde.org/show_bug.cgi?id=372116

5. Desktop notifications. We should fall back to notify-send if far2l extensions are not available

6. Cursor shape. There are escape seqs for modifying it:
https://www.freebsd.org/cgi/man.cgi?query=screen&apropos=0&sektion=4&manpath=FreeBSD+5.4-RELEASE&format=html
E[=s;eC
Set custom cursor shape, where s is the starting and e is the ending scanlines of the cursor.
Other options:
https://unix.stackexchange.com/questions/49485/escape-code-to-change-cursor-shape
https://github.com/kovidgoyal/kitty/issues/715
https://iterm2.com/documentation/2.1/documentation-escape-codes.html

7. BeginConsoleAdhocQuickEdit, GetLargestWindowSize, SetConsoleWindowMaximized — can be done by terminal emulators theirselves.

8. keydown/keyup events — no alfernative for terminal, I afraid. But whose are only used in far2l for F* key suggestions?

9. SetFKeyTitles — no alfernative for terminal, I afraid

elfmz:
It will never be just like, and these small lacks there and here will kill all taste of FAR:) FAR is good not cuz to its FAR, but cuz has many little yet things that makes life easier - spirit of (old good) Windows apps that was lost sometime/sometime :( Just few examples:
1 Complex hotkeys are not distinguishing left and right Ctrl - this is must have to use bookmarks fluently, also not all terminals works same (good/bad) and from my experience no one terminal supports really all possible combinations.
[...]

PR #1610, opened 2023-04-13
iterm2 rich keyboard protocol support

unxed:
NB: cmd+v shortcut seems not to work in rich kb mode, so we disable this mode for 2 seconds if cmd key press is caught. also right alt is mapped to right control as there is no right control on some mac keyboards, and right_cmd+number hotkeys are used by iterm2 itself. also sscanf works ok here even if some parts of sequence are omitted.


unxed:
Touch #1101
Touch #1575

PR #1612, opened 2023-04-15
СMake: correctly configure OpenSSL/libarchive on Mac

vitalyster:
* Outdated FindOpenSSL.cmake dropped, now using stock one


elfmz:
There're builds for still Ubuntu 16 and even 14 which's cmake doesnt have FindOpenSSL as i guess..

vitalyster:
Tested on Ubuntu 16.04 - it has cmake 3.5 and FindOpenSSL.cmake module is present and openssl is detected.

Ubuntu 14.04 - has CMake 2.8 which is not supported by far2l (3.0.2 is minimum required version). Even if someone install newer cmake - it will have FindOpenSSL.cmake there.

unxed:
Checked on 16.04, on which I'm building a portable builds. Everything is fine.

vitalyster:
@elfmz old paths does not have "/opt/homebrew" prefix, so previous solution will not help on m1 Mac. 
Addressed in https://github.com/elfmz/far2l/pull/1623

PR #1613, opened 2023-04-15
cmake: cleanup build tree to avoid changes to tracked files on configure stage

vitalyster:
When CMake configuration happens on working tree we have some unintended changes to tracked files.
This patch will help to make working tree clean after configuration stage:
* use .in suffix to `configure_file` file and ignore configured result files
* ignore  *askpass and *sudoapp output binaries
* remove incsrch/makefile which overwrites by CMake (on case-insensitive filesystems)


elfmz:
in general its recommended to use separate directory for build but not sources root, thats why '_build' is there in .gitignore, but ok

PR #1637, opened 2023-04-24
Info about: Select screen fragment & immediately to clipboard

akruphi:
This is a potentially dangerous because set new data to clipboard without any confirmation.

User should have info about this feature.


akruphi:
> Suggest to use shortest possible phrasing as this screen can be rather small and long message then will be truncated. `Mouse selection immediately copies fragment to clipboard` and move it to very end and make it yellow color as it works always, but not only in 'While typing command' and is really important

Thank you. I agree with any text correction because my PR was only for user information and I may wrote not perfect short phrase.

PR #1641, opened 2023-04-26
uadb almost a full-fledged plugin

m32:
links are converted to taget types
more implemented features from vfs api

unhide pluginmanager module

### DIFF
--- a/python/src/python.cpp
+++ b/python/src/python.cpp
@@ -105,6 +105,7 @@ protected:
         std::string syspath = "import sys";
         syspath += "\nsys.path.insert(1, '" + pluginPath + "')";
         syspath += "\nsys.path.insert(1, '" + pluginPath + "/plugins')";
+
         PyRun_SimpleString(syspath.c_str());
 
         PyObject *pName;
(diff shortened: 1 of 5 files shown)


python/src/python.cpp
```diff
@@ -383,13 +378,16 @@ XPORT(int, MakeDirectory)(HANDLE hPlugin,const wchar_t **Name,int OpMode) {
     PYTHON_INT(0)
     return result;
 }
+#if 0
+// in far2l only one plugin is allowed at time
```
elfmz, line 382: what was wrong with this?

m32: Currently OpenFilePlugin uses the plugin manager itself, not the plugins installed with it.
It's the same with the analyze api

I still have no idea how to use this api, so I turned it off.

elfmz: Yep its same as AnalyseW, but only difference that later is FAR3-specific.
A alternative usage example can tell that NetRocks uses it to 'open' special ini file that describes connection.

m32: For now, I will finish this plugin, in a moment there will be another pull anyway - adb is much faster :)

PR #1642, opened 2023-04-26
python adb

m32:
improvements required by the adb plugin, and the plugin itself works significantly faster
- busybox vs toolbox vs toybox vs pure adb api
- configuration stored in ini file


elfmz:
Please see: https://github.com/elfmz/far2l/actions/runs/4812265096/jobs/8567281192

elfmz:
image

PR #1643, opened 2023-04-26
adb undefined variable

m32:
I compile python separately and use ready python-venv, sorry.


elfmz:
may be then do #ifdef VIRTUAL_PYTHON_PATH  #else  ... #endif to make it possible both ways?

PR #1646, opened 2023-04-26
No redundant recompile on git revision change any more

unxed:
see https://github.com/elfmz/far2l/issues/1515#issuecomment-1451851021

NB! Copyright string was encoded using XOR since at least version 1.64 (see ctrlobj.cpp [in 1.64](https://github.com/FarGroup/FarManager/releases/tag/release%2F1.64)). Probably the reason was to prevent from removing or changing copyright string in binary file. We probably do not need such protection currently, so I do not reproduced it in new copyright string generation method.

### DIFF
--- a/far2l/CMakeLists.txt
+++ b/far2l/CMakeLists.txt
@@ -2,14 +2,31 @@ project(far2l)
 
 add_subdirectory (bootstrap)
 
+function(int_to_hex NUMBER HEX_STRING)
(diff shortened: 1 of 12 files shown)


alexax66:
Это так же  решило ранее описанную  проблему при сборке FAR2L на DragonFly BSD:
https://github.com/elfmz/far2l/pull/1594#issue-1659466557

PR #1647, opened 2023-04-27
fix #1634

### LINKED ISSUE #1634: spaces are inserted instead of tabs sometimes
1. open new file in editor
2. switch to T4 mode (do not expand tabs, tab size = 4)
3. press `Right Arrow` 4 times
4. press `i`
5. press `F5`
6. You will see 4 spaces inserted at the beginning of a line. But it should be 1 tab instead.

unxed:
fix #1634

insert tabs if possible instead of spaces if cursor is beyond EOL and char key is pressed (or paste happend) if tab mode is 0

### DIFF
--- a/far2l/src/editor.cpp
+++ b/far2l/src/editor.cpp
@@ -2528,36 +2528,63 @@ int Editor::ProcessKey(int Key)
 
 				if (Key < 0x10000 && CurPos > Length) {
 
-					//if (PrevLine) {
-						
-						int TabPos = CurLine->GetCellCurPos();
-						CurLine->SetCurPos(Length);
-
-						for (int I = Length; I < CurPos; I++) {
-
-							int NewTabPos = CurLine->GetCellCurPos();
-
-							if (NewTabPos == TabPos)
-								break;
-
-							if (NewTabPos > TabPos) {
-								CurLine->ProcessKey(KEY_BS);
-
-								while (CurLine->GetCellCurPos() < TabPos)
-									CurLine->ProcessKey(' ');
-
+					// detect space alignment by search for lines starting with space
+					Edit *PrevLine = CurLine->m_prev;
+					bool SpaceAligned = false;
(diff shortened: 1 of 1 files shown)


elfmz:
Not sure if its valid/expected behavior, why tabs should be inserted just due to cursor movement?

unxed:
Oh, it's not about inserting tabs on cursor movement. It's about situation when you start typing (or pasting) having cursor beyond end of a line. Old code used previous lines to detect if tabs or spaces should be used to fill the gap between end of line and cursor position. But previous line may have no tabs or spaces at all. More of this, previous line is not available if you are at the first line of a file. So new code uses tabs mode to detect if tabs or spaces should be used for filling the gap from end of line to cursor position. Also it fixes a weird bug when tab or space sometimes was inserted after cursor position and not before it as it should.

elfmz:
I'd say that this behavior makes sense and its same in FAR3. Note that setting is about what to do on pressing Tab key, but not about expanding anything by tabs.
I see however one inconsistency in the original behavior that if it uses same amount of tabs as previous line, causing mixed tab/space indentation to next line, like:
```
	some
		original
	lines
	            newly typed line
```
I'd suggest to keep original behavior if indentation way of other line(s) is known, and use TABs/Spaces setting only if its unknown. And use tabs to fill all space from position zero to beginning of line. HOWEVER if line being typed goes after another text, like
```
	some
		original
	lines
	existing          newly typed
```
then don't use tabs between 'existing' and 'newly typed'

unxed:
added two additional conditions:

- if space is found at the beginning of one of previous lines, do not use tabs for filling gaps from line end to cursor position
- if non-space char is found in the current line, do not use tabs for filling such gaps also

unxed:
забыл важное! вставка пробелов/табов в Undo не учитывается (и так до меня ещё было, кстати — противный баг).

надо поправить. #1650

PR #1650, opened 2023-04-27
include space/tabs insertions in Undo history

unxed:


elfmz:
I think need to declare feature freeze until 2.5.1, only bugfixes..

Issue #1592, opened 2023-04-07
can not change charset if .editorconfig file present


unxed:
Try to change file encoding of any file in far2l's source tree in editor using F8 or Shift-F8.

unxed:
Probably it's a feature, not a bug! accident charset change hurts! Probably we don't need fixing it, but just documenting it. Or, at lease, we may show some confirmation box before charset change in such cases. And, anyway, tab mode and size should behave the same way as charset.

akruphi:
PR with code of warnings was merged. This issue may be closed.

PR #1707, opened 2023-06-15
Flags to AddIntEditField() & Editor: warnings when trying to change values set in .editorconfig

akruphi:
* Add `Flags` to `AddIntEditField()`

* If charset, indent_style & indent_size are set in **.editorconfig** added warnings (text now hardcore in code) about the prohibition of change them in Editor by F8/Shift-F8, Ctrl-F5, Shift-F5 & disable them fields in local Editor settings by F9/Alt-Shift-F9

This PR explain situation #1592 for charset
and add simular behaivour for other .editorconfig parameters after 6ff9f77

I'm not sure for good "????" in `fileedit.cpp` line 1131. Please verify and may be correct message near "????" to more appropriate text.


elfmz:
??? is fine there, but need then move all message strings to farlang.templ.m4

akruphi:
> ??? is fine there, but need then move all message strings to farlang.templ.m4

Can `farlang.templ.m4` and/or function `Message()` operate with multi-line strings in one variable or only they work with 1-line strings?

If it work only with 1-line strings all news message will need a lot of variables in `farlang.templ.m4` for these messages :(

PR #1743, opened 2023-07-09
Set Attributes Dialog: after #1742 more details info if selected>1

akruphi:


elfmz:
need not to forget to add translations to new text strings

PR #1744, opened 2023-07-10
Set Attributes Dialog: add draft confirmation if new symlink target not exist

akruphi:
Now confirmation texts are draft, it will need to be written more accurately.


elfmz:
I think it would be better just to change color of edit control using DM_SETCOLOR message in realtime during editing (from DN_EDITCHANGE) than adding message box.

elfmz:
well, but dialog still can be useful even with such solution, so ok

PR #1756, opened 2023-07-17
setattr: Explicit Info for special files - more flexible after #1755

akruphi:


elfmz:
but won't this just duplicate same information?

elfmz:
Then may be its better to use file command info if it didn't fail? (need to modify PClose behavior to get exit code)

> Additional question: can now far2l in Attributes dialog after it initialization in Info field put cursor to 1st symbol (now always after last symbol and long text scrolling and it hide very informative start of string)?

If you will find how to do this - it would be fine..

akruphi:
> Then may be its better to use file command info if it didn't fail? (need to modify PClose behavior to get exit code)

May be. Good idea.

> > Additional question: can now far2l in Attributes dialog after it initialization in Info field put cursor to 1st symbol (now always after last symbol and long text scrolling and it hide very informative start of string)?
> 
> If you will find how to do this - it would be fine..

I'm not so familiar with far2l and now not know how do it.

PR #1770, opened 2023-07-25
fix minor issues in help files

unxed:


packaging/suse/far2l.spec
```diff
@@ -21,7 +21,7 @@ Distribution: 	openSUSE Tumbleweed
 BuildRequires:  cmake
 BuildRequires:  gcc-c++
 BuildRequires:  pkgconfig
-BuildRequires:  gawk m4 make
```
elfmz, line 24: isn't awk must be removed here but not m4?

unxed: .m4 files are parsed using far2l_m4 Perl script now.

unxed: It does not require m4 itself.

elfmz: ok, but still is gawk should be listed in BuildRequires here?

unxed: Now everything is done in Perl, neither awk nor m4 needed any more.

unxed: fixed

PR #1809, opened 2023-08-23
Update CMakeLists.txt to fix python plugin

atolismesh:
@m32
Update CMakeLists.txt to fix python plugin.   
Some key python plugin files copied to wrong destination by install procedures. Fix it by changing  "share" to "lib".
Without this fix command "py:load" says "py:load: command not found"


atolismesh:
@unxed

unxed:
currently my debs are built w/o python, but prs are welcome :)

PR #1815, opened 2023-08-26
python update

m32:
#pragma pack - in far2l it's different from winapi
handling editor events inside plugins

new plugins:
uedsort.py - sort selected block in editor (select text and use Python ED sort form F11 menu)
uedindent.py - indent/dedent selected block of text with TAB or SHIFT+TAB
uprogrss.py - sample progress dialog


akruphi:
@m32 thank you for :fire: py-plugins system.

About Tab/Shift-Tab indent/dedent in editor for row block it was fixed recently in build-in plugin [SimpleIndent](https://github.com/elfmz/far2l/commit/c452a20adeef5fd6a38f7a4288348e6b48f4c3ec) and now works properly in far2l.

PR #1862, opened 2023-09-26
clang-format update and schema overrides

ma5ter:
.clang-format updated to the current syntax according to the [ClangFormatStyleOptions](https://clang.llvm.org/docs/ClangFormatStyleOptions.html )

additionally schema overrides for CLion added as .clang-format.json to reflect changes from https://github.com/elfmz/llvm-project repo


unxed:
Do I understand you correctly that these changes will allow us to achieve the result we need on the vanilla version of clang?

Issue #1865, opened 2023-09-27
DragonFly BSD: Far2l hangs when entering any  command in console after commit 07f1ad3


alexax66:
Far2l hangs when entering command in console after commit 07f1ad3 on DragonFly BSD.
When I[ reverted](https://github.com/alexax66/far2l/commit/21424c3f00e11014031fa4fbb13dfb441cbabdf7)  commit 07f1ad3  all works fine again.

PR #1867, opened 2023-09-27
Adapt tsset/getattr signal handling for DragonFly BSD

alexax66:
This fix issue  #1865 for DragonFly BSD


elfmz:
Thanks but that change was 'just in case' and if it causes problems anywhere then its really better to revert it and just add some logging instead

alexax66:
Last commits fixed this problem. Thanks.

PR #1878, opened 2023-10-03
view sh - use more coloring tools and use sed instead of perl or awk

sxored:


sxored:
thanks for #1672 and #1772

PR #1930, opened 2023-12-08
Fix incorrect reload in Editor & minor Editor/viewer codepage in status bar tuning

### LINKED ISSUE #1918: Switching encoding breaks syntax highlighting
When switching from 1-byte encoding to utf-8 in editor, it jumps scroll to the beginning of the file and breaks syntax highlight:

изображение

Happens both in GUI and in TTY

akruphi:
Also fix #1918


unxed:
#1918 is indeed very annoying

PR #1981, opened 2024-01-27
Add missing ssh port in ways.ini

atolismesh:
Add missing ssh port in ways.ini. 
Without this fix NetRocks Shell protocol ignores connection port in configuration and unable to connect to nonstandard port.


atolismesh:
To install this fix correctly, you must completely remove the _build/install folder and rebuild Far2l.

PR #2020, opened 2024-02-22
Miscellaneous commits

akruphi:


CMakeLists.txt
```diff
@@ -45,10 +45,11 @@ endif()
 
 if(NOT "${GIT_TAG}" STREQUAL "v_${VERSION}")
     if(GIT_FOUND)
-        execute_process(COMMAND "${GIT_EXECUTABLE}" show --no-patch --format=%cs HEAD
+        execute_process(COMMAND "${GIT_EXECUTABLE}" show --no-patch --format=%ci HEAD
             WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
-            OUTPUT_VARIABLE GIT_DATE
+            OUTPUT_VARIABLE GIT_DATE_FULL
             OUTPUT_STRIP_TRAILING_WHITESPACE)
```
elfmz, line 51: but why later GIT_DATE still being used?

akruphi: In this method full date via %ci obtained in GIT_DATE_FULL, then in SUBSTRING it cutt in GIT_DATE

shmuz:
This patch uses `GIT_DATE_FULL` as a temporary variable and then makes `GIT_DATE` out of it.
Probably `GIT_DATE` only could be used, as below (not tested):
```cmake
            OUTPUT_VARIABLE GIT_DATE
            OUTPUT_STRIP_TRAILING_WHITESPACE)
        string(SUBSTRING ${GIT_DATE} 0 10 GIT_DATE)
```

akruphi:
@elfmz: far2l has hang when try open help section `@CmdLineCmd` in Ukrainian language. An these hang was before changes in PR.

CMakeLists.txt
akruphi: Add commit with use only one variable. But I has not systems with old git and can't test in it.

PR #2023, opened 2024-02-26
Increased size of mark in file highlighting

anta999:


elfmz:
can you attach screen shots to quickly see what changed visually?

Issue #1889, opened 2023-10-18
interface is disturbed in WezTerm


unxed:
1. run
far2l --tty --nodetect=x
in [WezTerm](https://github.com/wez/wezterm)

2. press Ctrl+O two times

3. open any short file in viewer or editor

unxed:
![изображение](https://github.com/elfmz/far2l/assets/1151423/ec69bff1-825d-47ce-a44f-e0b5928912e9)

sirburpalot:
Same problem here. The UI looks normal immediately after launching far2l, but becomes corrupted after performing any action that clears the screen, like pressing Ctrl+O or viewing a file using F3.

![image](https://github.com/elfmz/far2l/assets/1429516/d67d84b7-f140-4ef0-9c52-6cd62e60f47a)

unxed:
I discovered an interesting pattern. The interface seems to break when trying to print a line off the screen. Like here (far2l after ctrl+o in gnome terminal and wezterm, see `Copyri`

![Снимок экрана от 2024-03-20 21-57-25](https://github.com/elfmz/far2l/assets/1151423/25d4312d-4774-45ef-8133-be36cdfa66b1)

unxed:
Also please note the position of ArrowUp char. It should be printed int the end of a line, but in WezTerm it is in the beginning

![изображение](https://github.com/elfmz/far2l/assets/1151423/69711ac1-f0b8-4797-ae61-4f3065cf93e2)

unxed:
Implemented workaround on our side:
https://github.com/elfmz/far2l/pull/2089/files

Issue #2084, opened 2024-03-19
viewing binary files breaks output to kitty terminal


unxed:
Noticed [here](https://habr.com/ru/articles/593451/comments/#comment_23787091):

> When viewing binary files, `far2l` consistently blows the minds of the `kitty` terminal (it starts drawing weird things). On the one hand, the terminal itself probably has bugs, but `mc` under similar conditions does not produce such an effect.

And [here](https://t.me/far2l_ru/15957):

> visual artifacts and audio beeps

Kitty has this behavior [documented](https://sw.kovidgoyal.net/kitty/faq/#i-catted-a-binary-file-and-now-kitty-is-hung) and suggests [using](https://unix.stackexchange.com/a/522300/422943) `cat -v` for binary data:
> I catted a binary file and now kitty is hung?
> 
> Never output unknown binary data directly into a terminal.
> 
> Terminals have a single channel for both data and control. Certain bytes are control codes. Some of these control codes are of arbitrary length, so if the binary data you output into the terminal happens to contain the starting sequence for one of these control codes, the terminal will hang waiting for the closing sequence. Press [ctrl+shift+delete](https://sw.kovidgoyal.net/kitty/conf/#shortcut-kitty.Reset-the-terminal) to reset the terminal.
> 
> If you do want to cat unknown data, use cat -v.

mc does not have this bug, it just replaces all control codes with `.` char.

How to reproduce?
1. `sudo apt install kitty`
2. run `kitty`
3. `far2l --tty` in kitty
4. view `initrd.img`
5. hold `Down` key
6. wait for beeps

akruphi:
Dupe #2068

unxed:
Under kitty? Kovidgoyal's kitty?

unxed:
Please note that even with `-a` option beeps are still happening. Looks like `-a` do not work as expected.

unxed:
@Dazzar56 was planning to look into this.

unxed:
Actually, it's three separate issues.

1) Some terminals produce sound on BELL char, so we probably should not output it.

2) Visual artifacts in TTY mode (tested in GNOME terminal and kitty). To reproduce:
A) view this file as UTF-8, wrap disabled
https://upload.wikimedia.org/wikipedia/commons/a/af/K%C3%B6lner_Dom_-_Westfassade_2022_ohne_Ger%C3%BCst-0968.jpg
B) hold Down key
C) look at the bottom of the screen
D) be patient
 
[...]

PR #2089, opened 2024-03-20
disable lazy cursor operations in WezTerm as it does not support them

unxed:
workaround for #1889


sirburpalot:
Just checked this, and it does solve the problem.

unxed:
Looks like some other terminals, for example, kitty, also suffer from this problem:
https://github.com/elfmz/far2l/issues/2084#issuecomment-2016003044

PR #2105, opened 2024-03-24
fixes artifacts in viewer in tty mode

### LINKED ISSUE #2084: viewing binary files breaks output to kitty terminal

unxed:
fixes #2084


unxed:
CharClasses.cpp regenerated on Ubuntu 23.10

unxed:
For example. Run far2l in kovidgoyal's kitty, open far2l's binary in viewer. You will see strange artifact in the right column. This PR fixes that.

Снимок экрана от 2024-03-25 17-14-14

PR #2131, opened 2024-04-08
Added tests for CharClasses.cpp

unxed:
Added simple test showing differences between char width detected by wcwidth() and by CharClasses.cpp functions


unxed:
Added one more similar test, using https://github.com/JuliaStrings/utf8proc library that may be more reliable. Anyway, results can not be trusted blindly, output of such tests is just a starting point for further research.

unxed:
The third test should give the most accurate results, since it uses a character width detection code generated by parsing the latest data downloaded from unicode.org

PR #2152, opened 2024-04-12
Colortest2

anta999:
Color dialog updated
Mark inherit
Add color masks

some fix


akruphi:
After this PR new defaults for **Files highlighting and sort groups** not good: some files has not left mark, but other has mark symbol and in panel **left real symbol of filename** not always **aligned uniformly**. Please either add even a space mark to everyone (it may be difficult because marks can additive), or remove any default mark symbols.

PR #2154, opened 2024-04-13
fix #2139: type of nodetect variable changed from char* to dword (now used as bitmask)

### LINKED ISSUE #2139: --nodetect=xi does not work from FAR2L_ARGS
Just `--nodetect` still do.

spnethw:
fix #2139:

Method ArgOptions::ParseArg() [used](https://github.com/elfmz/far2l/blob/e6ad61a1ebfd816232f24bf5e224bf7ef83c2ebc/WinPort/src/Backend/WinPortMain.cpp#L288) internal buffer of [str](https://github.com/elfmz/far2l/blob/e6ad61a1ebfd816232f24bf5e224bf7ef83c2ebc/WinPort/src/Backend/WinPortMain.cpp#L349) (**local** std::string) [passed](https://github.com/elfmz/far2l/blob/e6ad61a1ebfd816232f24bf5e224bf7ef83c2ebc/WinPort/src/Backend/WinPortMain.cpp#L355) from WinPortMain(), this led to the corruption of the variable value.

Type of nodetect variable  was changed from char* to DWORD, and now it is used as a bitmask.

### DIFF
--- a/WinPort/src/Backend/Backend.h
+++ b/WinPort/src/Backend/Backend.h
@@ -8,7 +8,7 @@
 ///   Something changed in code below.
 ///   "WinCompat.h" changed in a way affecting code below.
 ///   Behavior of backend's code changed in incompatible way.
-#define FAR2L_BACKEND_ABI_VERSION	0x08
+#define FAR2L_BACKEND_ABI_VERSION	0x09
 
 #define NODETECT_NONE   0x0000
 #define NODETECT_XI     0x0001
--- a/WinPort/src/Backend/Backend.h
+++ b/WinPort/src/Backend/Backend.h
@@ -10,6 +10,12 @@
 ///   Behavior of backend's code changed in incompatible way.
 #define FAR2L_BACKEND_ABI_VERSION	0x08
 
+#define NODETECT_NONE   0x0000
+#define NODETECT_XI     0x0001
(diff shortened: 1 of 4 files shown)


WinPort/src/Backend/WinPortMain.cpp
```diff
@@ -288,11 +288,18 @@ struct ArgOptions
...
 		} else if (strstr(a, "--nodetect=") == a) {
-			nodetect = a + 11;
-
+			if(strstr(a+11,"xi")) {
+				nodetect = NODETECT_XI;
+			} else if (strchr(a+11,'x')) {
+				nodetect = NODETECT_X;
+			}
+			if(strchr(a+11,'f'))
+			{
```
elfmz, line 300: please use consistent styling (egyptian brackets in this case)

WinPort/src/Backend/Backend.h
```diff
@@ -10,6 +10,12 @@
 ///   Behavior of backend's code changed in incompatible way.
 #define FAR2L_BACKEND_ABI_VERSION	0x08
```
elfmz, line 11: increase version (to guard against wrong version of far2l_gui.so )

spnethw:
fixed

PR #2285, opened 2024-07-14
fix Plugin Colorer crashes #2277

ctapmex:
init main variable on startup

ps. but I still don't understand why after unloading and loading the plugin the "old" memory was used


elfmz:
In POSIX dlclose doesnt guarantee that library will be really unloaded: https://stackoverflow.com/questions/24467404/dlclose-doesnt-really-unload-shared-object-no-matter-how-many-times-it-is-call

Issue #2306, opened 2024-07-26
linking fails


unxed:
```
l but the CMakeFiles/NetRocks-FTP.dir/src/Protocol/FTP/FTPConnection.cpp.o: In function `OpenSSLContext::sNewClientSessionCB(ssl_st*, ssl_session_st*)':
FTPConnection.cpp:(.text._ZN14OpenSSLContext19sNewClientSessionCBEP6ssl_stP14ssl_session_st+0x34): undefined reference to `SSL_SESSION_up_ref'
CMakeFiles/NetRocks-FTP.dir/src/Protocol/FTP/FTPConnection.cpp.o: In function `OpenSSLContext::OpenSSLContext(StringConfig const&)':
FTPConnection.cpp:(.text._ZN14OpenSSLContextC2ERK12StringConfig+0x40): undefined reference to `TLS_client_method'
```

https://github.com/elfmz/far2l/issues/1985#issuecomment-2253224909

unxed:
> I tried to build it on Mint 20.3 and CentOS8 stream.

ctapmex:
no problem
mint
```dockerfile
FROM linuxmintd/mint20.3-amd64

SHELL ["/bin/bash", "-xeuo", "pipefail", "-c"]

# disable auto-cleaning of the apt cache, so that docker caching works
#RUN rm -f /etc/apt/apt.conf.d/docker-clean

# use docker cache for apt packages
RUN --mount=type=cache,target=/var/cache/apt \
<<EOT
    apt-get update && apt-get upgrade -y
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends cmake gcc g++ make \
     libuchardet-dev libxerces-c-dev libwxgtk3.0-gtk3-dev libx11-dev libxi-dev libssl-dev libsmbclient-dev libnfs-dev libneon27-dev libssh-dev  \
     libarchive-dev libpcre3-dev	        
    rm -rf /var/lib/apt/lists/*
EOT

WORKDIR /code
RUN <<EOT
    git config --global http.sslVerify false 
    git clone --depth 1 https://github.com/elfmz/far2l.git
    mkdir build
    cmake -DUSEWX=YES -DCMAKE_BUILD_TYPE=Release  -B /code/build -S /code/far2l
    cmake --build /code/build --config Release -j$(getconf _NPROCESSORS_ONLN)
EOT

```

centos8 stream
```dockerfile
FROM tgagor/centos-stream:stream8

[...]

PR #2308, opened 2024-07-28
GitHub Action: clean workflow runs

ctapmex:
Clean old workflow runs logs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 name: check build
 
 on:
-  # allow manually running worflow, useful for testing branches
+  # allow manually running workflow, useful for testing branches
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


elfmz:
Why removed openssl? if its related to #2306 then its not right thing to do - need to fix openssl as its required for FTPS support, without installed openssl FTPS will not be supported

ctapmex:
These are the instructions for building on github. Some libraries have already been installed there. Look at the comment "#libx11, openssl is already installed".

if this is not corrected, then there are warnings in the build logs

[macos14-arm64-clang15](https://github.com/elfmz/far2l/actions/runs/10131553973/job/28014417776#step:3:10)
libx11 1.8.9 is already installed and up-to-date. To reinstall 1.8.9, run: brew reinstall libx11

PR #2328, opened 2024-08-02
Fix CheckInsidePlugin for non-panel plugins

rampitec:
We are checking if we are inside a plugin by testing CurPluginItem. This does not have a value until OpenPlugin returns a handle. It is a common practice to perform all work inside a non-panel, non-persistent plugin within its OpenPlugin call and return INVALID_HANDLE_VALUE.

There is PSIF_ENTERTOOPENPLUGIN flag, but we do not set it. The fix is to set it and check it.

Fixes https://github.com/elfmz/far2l/issues/2325, but it also needs a change in the ~/.config/far2l/settings/key_macros.ini to include NoSendKeysToPlugins=0x1 which has replaced NoInsidePlugin back in
2003. Added incsrch/key_macros.ini for convenience to add to the key_macros.ini. Please refer to bug #2325 for in deep details.

### DIFF
--- a/far2l/src/macro/macro.cpp
+++ b/far2l/src/macro/macro.cpp
@@ -6440,7 +6440,9 @@ BOOL KeyMacro::CheckEditSelected(DWORD CurFlags)
 
 BOOL KeyMacro::CheckInsidePlugin(DWORD CurFlags)
 {
-	if (CtrlObject && CtrlObject->Plugins.CurPluginItem && (CurFlags & MFLAGS_NOSENDKEYSTOPLUGINS))		// ?????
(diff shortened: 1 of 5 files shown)


rampitec:
@unxed @shmuz plz review if you can.

unxed:
I have no objections. The only question is: can a plugin be launched from a plugin? I'm not sure about that.

shmuz:
> @unxed @shmuz plz review if you can.

No objections.

rampitec:
> I have no objections. The only question is: can a plugin be launched from a plugin? I'm not sure about that.

I do not see a way for plugin code to call PluginManager::OpenPlugin(). I don't believe this is supported scenario.

PR #2335, opened 2024-08-05
Incsrch: add ^g/^#g as a duplicate of ^Enter/^#Enter

rampitec:
Some terminals (like the default terminal of Ubuntu 24.04) steal
^Enter from the apps. Adding duplicate hotkeys to work in --tty
mode.

Bumped version to 2.2.


rampitec:
Includes changes from https://github.com/elfmz/far2l/pull/2328 as there is no easy way to create stacked reviews on github.

rampitec:
@unxed The change we have discussed.

incsrch/control.c
```diff
@@ -80,6 +80,7 @@ BOOL CollectEvents(void)
 							PutEvent(KC_BACK);
 							continue;
 						case VK_RETURN:
+						case 'G':
```
rampitec, line 83: That said: this is the only functional change.

elfmz:
again want to remind that there is 'sticky controls ' feature, in this case Pressing Ctrl+Space then Enter should work as Ctrl+Enter

rampitec:
> again want to remind that there is 'sticky controls ' feature, in this case Pressing Ctrl+Space then Enter should work as Ctrl+Enter

Do you mean in the gnome-terminal? Anyway, pressing 2 combinations is less convenient.

rampitec:
I will also have to rebase this on https://github.com/elfmz/far2l/pull/2338 or that on this, whetever and if merged first into master.

rampitec:
> again want to remind that there is 'sticky controls ' feature, in this case Pressing Ctrl+Space then Enter should work as Ctrl+Enter

I have booted Ubuntu and tried sticky keys. It is actually not convenient, I really need to press it twice, and usually when searching it takes multiple 'find next' attempts, while ^space only lasts for one single keypress.

I also have to clarify, this is not a global hotkey. It is only when plugin reads input within its own OpenPlugin.

rampitec:
Ping

rampitec:
Ping. I am using it for a month or so w/o new issues.

PR #2338, opened 2024-08-09
[incsrch] Cleanup windows stuff

rampitec:
Remove files not directly related to far2l build.
Add Belorussion lng to cmake, thanks for contribution.


elfmz:
But why incsrch/gpl.txt also removed?

rampitec:
> But why incsrch/gpl.txt also removed?

I just noticed this is the only plugin having it. Do you want me to restore it?

elfmz:
yep its a kind of things to be better kept unless negotiated with original author(s)

rampitec:
> yep its a kind of things to be better kept unless negotiated with original author(s)

Well, I am the original author ;) But I am OK anyway.

elfmz:
ah:) then its (would be) ok

rampitec:
> ah:) then its (would be) ok

Do you want me to add a separate change to remove it? I really do no want it to differ from general far2l licensing. The purpose it's there was to say: this stuff is free. I think far2l already has a copy for GPL license, although somewhere in the debian packaging, but it is still there. I am OK either way, just want to minimize source clutter.

PR #2358, opened 2024-08-30
Update Findfmt.cmake for compatibility with fmt 11.x.x

atolismesh:
After upgrading from fmt 10.x.x to 11.x.x in openSUSE, compilation of the colorer plugin caused cmake errors.   fmt 11.x.x stores version in new file fmt/base.h.
Update Findfmt.cmake to be compatible with fmt 11.x.x

### DIFF
--- a/cmake/modules/Findfmt.cmake
+++ b/cmake/modules/Findfmt.cmake
@@ -1,7 +1,9 @@
 find_path(fmt_INCLUDE_DIR NAMES fmt/format.h)
 
-if(fmt_INCLUDE_DIR)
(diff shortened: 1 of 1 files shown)


atolismesh:
@ctapmex 
ping

ctapmex:
fmt (as part of spylog) and spdlog are optional to use. By default, they are disabled for use.
but anyway, I'm already in the process of removing spdlog from the plugin completely. These dependencies will disappear from cmake.

atolismesh:
Thanks!

elfmz:
fmt always being causing problems due to its changes across versions and even across different config flavors in different repos

PR #2380, opened 2024-09-08
Add align filenames by marks

anta999:
Для любителей MC.
ALIGN

### DIFF
--- a/far2l/src/hilight.cpp
+++ b/far2l/src/hilight.cpp
@@ -470,31 +470,31 @@ class HighlightFilesChunk : protected Threaded
 	uint64_t _CurrentTime;
 	const TPointerArray<FileFilterParams> &_HiData;
 	FileListItem **_FileItem;
-	int _FileCount;
+	size_t _FileCount, _LongestMarkLength;
 	bool _UseAttrHighlighting;
 
 	virtual void *ThreadProc()
 	{
-		DoNow();
-		return nullptr;
+		return DoNow();
 	}
 
 public:
 	HighlightFilesChunk(uint64_t CurrentTime, const TPointerArray<FileFilterParams> &HiData,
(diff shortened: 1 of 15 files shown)


far2l/src/hilight.cpp
```diff
@@ -1029,32 +1063,15 @@ void HighlightFiles::SaveHiData()
...
-		for (size_t j = 0; j < ARRAYSIZE(hl1.Color[i]); ++j) {
-			if (hl1.Color[i][j] != hl2.Color[i][j])
-				return false;
-			if (hl1.Mask[i][j] != hl2.Mask[i][j])
-				return false;
-		}
-	}
-
-	return true;
+	return !memcmp(&hl1, &hl2, sizeof(HighlightDataColor));
```
elfmz, line 1066: Then may be use simple (default) hl1 == hl2? memcmp has potential risk of wrong result if structure has alignment gaps between fields

elfmz: what was wrong with original approach? may replace that long loop with hl1.Color == hl2.Color but everything else better leave in C++ style (i.e. with explicit compares)

elfmz: Я попозже впомичу вариант (ща немножечко нет возможности вкомитить), как я имел ввиду в коментах выше:
```
static bool operator==(const HighlightDataColor &hl1, const HighlightDataColor &hl2)
{
	if (hl1.Flags != hl2.Flags || hl1.MarkLen != hl2.MarkLen)
		return false;

	if (hl1.MarkLen && wmemcmp(&hl1.Mark[0], &hl2.Mark[0], hl1.MarkLen) != 0)
		return false;

	return (hl1.Color == hl2.Color && hl1.Mask == hl2.Mask);
}

```

PR #2440, opened 2024-10-14
fix memory leak in colorer

ctapmex:
fix memory leak in colorer


ctapmex:
included in https://github.com/elfmz/far2l/pull/2445

Issue #2387, opened 2024-09-17
How to access console output from Far Plugin?


faerot:
Is it possible? How?

akruphi:
Far2 API documentation (from classic windows version) you can see in https://api.farmanager.com/ru/index.html (in Russian).
far2l API changes you can see in https://github.com/elfmz/far2l/blob/master/HACKING.md

May be `SaveScreen` and `Text` from https://api.farmanager.com/ru/miscapi/index.html help you.

shmuz:
@akruphi Your 1-st link is for Far3 API not Far2
@faerot I can send you Far2 Encyclopedia (contains API) in CHM format

akruphi:
Far2 Encyclopedia CHM can be downloading within far2.x86.msi/far2.x64.msi from https://farmanager.com/history.php (two last links)

faerot:
Sorry, probably I was not clear enough. I am not talking about current screen of Far, I am talking about terminal output of the last command. The one that is covered by panels, SaveScreen will not help me here. Didn't find anything in the encyclopedia.

elfmz:
And may be it can be useful for you - its how console history is accessible inside far2l itself: https://github.com/elfmz/far2l/blob/master/far2l/src/vt/vtlog.h

faerot:
I assume
```cpp
std::string GetAsFile(HANDLE con_hnd, bool colored, bool append_screen_lines = true);
```
Is what would be helpful if exposed to plugin.

m32:
A much better solution is to make WinPort.h functions available to Python, just like C/C++ plugins have.

and later:
# switch to user screen
self.parent.info.Control(self.hplugin, self.ffic.FCTL_GETUSERSCREEN, 0, 0)
hr = winport.GetConsoleScreenBufferInfo(...)
hr = winport.ReadConsoleOutput(...)
# switch back
self.parent.info.Control(self.hplugin, self.ffic.FCTL_SETUSERSCREEN, 0, 0)

faerot:
> A much better solution is to make WinPort.h functions available to Python, just like C/C++ plugins have.
> 
> switch to user screen: self.parent.info.Control(self.hplugin, self.ffic.FCTL_GETUSERSCREEN, 0, 0)
> 
> hr = winport.GetConsoleScreenBufferInfo(...) hr = winport.ReadConsoleOutput(...)
> 
[...]

PR #2454, opened 2024-10-20
added api allowing to export VT history

elfmz:
for https://github.com/elfmz/far2l/issues/2387


m32:
1. Why self.info.FSF.VTEnumBackground(self.ffi.NULL, 0) is always 0 ?
How to make it non 0 ?

2. each #defined identifier is expanded by the preprocessor and later needs to be redefined for python, for me it is more convenient to use enum type:

enum VTH_LOGEXPORT {
VTH_LOG_COLORED = 0x00000001,
VTH_LOG_WITH_SCREENLINES = 0x00000002
};

3. BOOL farAPIVTLogExportW(HANDLE con_hnd, DWORD vth_flags, const wchar_t *file)
file argument should be mandatory and in no case overwritten by VTLogExport, or even better it should be a handle to the file to be written to (int fd instead of const wchar_t *file)

elfmz:
1. put any pending terminal command to background with Ctrl+Alt+Z 
2. enum is possible, can change to it if it makes life easier, but IMHO its a bit abusive way of using enums :)
3. it overriten only if its an empty string, if its not empty - its not changed

Issue #2124, opened 2024-04-03
strange fastfind behavior 


unxed:
1. Install two keyboard layouts:
Russian
English
(order is important!)

2. Select English layout

3. Go to / in far2l wx

4. Press `Alt`+`r`

5. You will see `к` in the fast find field.

Reproduces on Mint 21.3 Cinnamon (x11). Not reproduces on Ubuntu 23.10 (Wayland).

Upstream issue:
https://github.com/wxWidgets/wxWidgets/issues/23421

unxed:
Touch #1374

PR #2473, opened 2024-10-28
Do not apply Alt+letter workaround under Wayland

unxed:
As it is reported in #2124, wxWidgets/wxWidgets#23421 is not reproduced under Wayland. So let us do not apply workaround on Wayland systems to minimize risk of possible misbehavior.


elfmz:
suggest also to exclude macos

unxed:
done!

Issue #2472, opened 2024-10-28
Replacing top on Ctrl+W ?


faerot:
I am using btop instead of top and would like to have it invoked on Ctrl+W.
I know I can easily replace it of panels via macro, but it is not obvious how to replace it in editor.
Perhaps there is some kind setting I am missing? Maybe it is worth adding option to system settings as there are a lot of "top" alternatives - htop, btop and different people would like to use different ones?

akruphi:
Now you can put in your home file `. ~/.config/far2l/ps.sh` with call any other program inside. It file use to override default from `$FARHOME/ps.sh`

For more flexible changing you can use macros.
For me it is preferable external far2l plugin from https://github.com/VPROFi/processes and now I use macros for call it by override **Ctrl-W**.
![image](https://github.com/user-attachments/assets/98ba0d72-f022-43f9-b800-fa4fef3deafd)

faerot:
I have created ~/.config/far2l/ps.sh with the following contents:
```
#!/bin/bash
btop
```
Now btop runs on Ctrl+W, but after quitting, it also launches top, so it is sort of runs that file in addition to just "top" not instead of it

akruphi:
Ooops!!! It is incorrect processing in $FARHOME/ps.sh after call user file.
If possible, you may make corrections in https://github.com/elfmz/far2l/blob/master/far2l/bootstrap/ps.sh

elfmz:
why incorrect? you can just write exit 0 in FARHOME/ps.sh

faerot:
Is there any use case where FARHOME/ps.sh would not exit and it would be a correct behavior? Cause I can't come up with one and the requirement to write exit in FARHOME/ps.sh only causes confusion

elfmz:
User may have his own ps that implements different behavior under some circostances, for instance when far2l runs in ssh session, but not locally. Anyway, if writing some code reduces possible usage scenarios without adding more, then its better not to write such code:)

faerot:
If you insist that user-made ps.sh should include exit command, then this should be documented in bootstrap/ps.sh at least and maybe creating dummy empty FARHOME/ps.sh on install with contents similar to:
```
#!/bin/bash
# You may override default top/htop with other command here. It should end with "exit 0", otherwise it will invoke top/htop afterwards
exit 0
```
[...]

PR #2474, opened 2024-10-28
Fixed issue #2472

faerot:
fixed: htop/top still called after ps.sh


elfmz:
not suitable from architecture point of view, however explanation was added to ps.sh regarding need for exit 0

PR #2525, opened 2024-11-23
Removed PCRE library support

unxed:
There are several reasons for that.

1. 7zip 23.01 supports RAW format itself, we do not need `fls` and `icat` for it now

2. Two different regexp engines (PCRE and Far internal one) in one app is abmigious and confuse users.

3. `custom.ini` is here since 2016, and the only use for it is actually wrapping around 7zip. More of this, nowadays we can just link to `7z.so` in distros and use 7zip as a library.

4. Any parsing of archiver output can be done directly in the command used to invoke it, using tools like awk or perl. You can then provide multiarc with, say, CSV format, which can be parsed using the standard multiarc method.

5. Minus one unneeded dependency.


elfmz:
additionally it makes sense to export regexp functionality from far2l to plugins

elfmz:
ой а там уже есть RegExpControl :)

PR #2597, opened 2024-12-26
Rename desktop files, so that far2l.desktop is the WX version

### LINKED ISSUE #2596: latest version runs in console (instead of running in GUI)
I have Ubuntu 22.04

In don't know what you've done in the latest version, but here is what I have:
image
image

What helps:   to get one of the previous versions, rebuild and install.

mitya57:
When an application does not set its own `.desktop` file name via X11 or Wayland mechanisms (which is not supported by wxWidgets yet, see wxWidgets/wxWidgets#24668), the desktop environment may rely on `argv[0]` to guess which desktop file is used.

So, when user runs the GUI version of far2l (no matter from the menu or manually), and then pins the current window to the panel, the DE will use `far2l.desktop`. So, `far2l.desktop` should be the GUI version. This is what it was before #2573, anyway.

So I renamed:
- far2l.desktop → far2l-tty.desktop
- far2ledit.desktop → far2ledit-tty.desktop
- far2l-wx.desktop → far2l.desktop
- far2ledit-wx.desktop → far2ledit.desktop

GitHub does not recognize renaming when you look at PR diff, but if you look at per-commit diff, it will show the renaming properly.

This also fixes #2596.


vladikoles:
I join this request.

Ubuntu 24.04 wayland:
The DE for far2l (WX GUI) is broken.
I also found:
When far2l (WX GUI) is running, the "New Window" action for the icon on the launcher leads to the launch of far2l terminal.
After renaming the file /usr/share/applications/far2l.desktop to /usr/share/applications/far2l.desktop.bak:
When far2l (WX GUI) is launched, its normal icon in the launcher not showing.
The context menu is shown without the "New Window" item.

PR #2643, opened 2025-01-21
Allow using Ctrl+' as an alternative to Ctrl+`

unxed:
There are several reasons for that:
1. Ctrl+` do not work in plain tty mode
2. Ctrl+` also do not work in xterm even with ModifyOtherKeys=1
3. Ctrl+' is intuitive replacement for Ctrl+` that can be easily remembered
4. Changing folder to ~ is one of the most frequently needed actions. It should be available in any mode with any backend


unxed:
Эта штука в голом tty всё же не работает, где-то я ошибся когда тестировал. Надо откатить, наверное

PR #2649, opened 2025-01-26
better kitty protocol support

unxed:


far2l/src/vt/vtshell.cpp
```diff
@@ -829,10 +832,11 @@ class VTShell : VTOutputReader::IProcessor, VTInputReader::IProcessor, IVTShell
 			}
 
 			if (!KeyEvent.bKeyDown)
-				return std::string();
```
elfmz, line 832: hm, why? theoretically return std::string() is more efficient than return std::string("")

unxed: fixed

PR #2703, opened 2025-03-30
changelog.md: Set release date for v2.6.5

mitya57:
Next time, please do it before the release. This way the generated `.metainfo` file in the release tarball will have an entry for the release.


elfmz:
There is a script used to make releate tag: https://github.com/elfmz/far2l/blob/master/packaging/version_bump_patch.sh
Ideally it also can modify changelog too

PR #2729, opened 2025-04-24
NetRocks: add support of libssh SSH_OPTIONS_PROXYCOMMAND option

lieff:
Recently I have issues with NetRocks/scp/sftp connects via jump hosts.
But I was able work around this using libssh SSH_OPTIONS_PROXYCOMMAND option (looks like "ssh -W [host]:22 jump_host") which do the jump.
So this MR adds SSH_OPTIONS_PROXYCOMMAND option support.


akruphi:
Please also brief describe new feature in section "Master (current development)" in https://github.com/elfmz/far2l/blob/master/changelog.md may be as
```
* _NetRocks plugin_: Add support of libssh SSH_OPTIONS_PROXYCOMMAND option
```

PR #2735, opened 2025-05-03
NetRocks: update chengelog

lieff:
As requested in https://github.com/elfmz/far2l/pull/2729


elfmz:
..давненько не видели)

PR #2736, opened 2025-05-03
+ Added 'z' command prompt variable that returns the "{current git br…

ujos:
This commit introduces the support of `Z` prompt variable that expands
* into the `{git_current_branch} ` string if the path of the CURRENT panel points to the GIT repository and 
* into the empty string otherwise


akruphi:
Good idea!

But may be do it **without use any additional library** during compilation and use way a la https://github.com/elfmz/far2l/blob/1c91cfa5362fe2a62446ac6ae10018508bb6ff90/far2l/src/panels/infolist.cpp#L789-L816

For example (this draft code, please test it before use):
```cpp
FARString GetGitBranchName(FARString const& path)
{
	FARString strDir = path;
	do {
		FARString strGit = strDir + L"/.git";
		struct stat s;
		if (stat(strGit.GetMB().c_str(), &s) == 0) {
			fprintf(stderr, "GIT: %ls\n", strGit.CPtr());
			std::string branch_name;
			std::string cmd = "git -C \"";
			cmd+= EscapeCmdStr(Wide2MB(strDir.CPtr()));
			cmd+= "\" branch --no-color --show-current";

			if (POpen(branch_name, cmd.c_str()) && branch_name.length() > 0) {
			       FARString ret = "{" + branch_name.c_str() + "}";
                               return ret;
			}
			break;
		}
	} while (CutToSlash(strDir, true));
	return {}:
}
```

ujos:
@akruphi Dunno... command line format sometimes change. libgit2 library seems does not invoke `git` executable.

akruphi:
> @akruphi Dunno... command line format sometimes change. libgit2 library seems does not invoke `git` executable.

If you build far2l yourself additional dependencies is not problem and result may be more robust.

But in repositories (now far2l in Debian / Ubuntu) maintainers try make packages only with essentials dependencies and far2l will build without linking to libgit2 because far2l should work in minimal systems.

ujos:
@akruphi Ubuntu repo contains the `libgit2` package. If this dependency causes issues, ok then, let's decline this PR.

akruphi:
> @akruphi Ubuntu repo contains the `libgit2` package. 

Yes, but not in all systems `libgit2` package installed.

You can add my variant of code (after tuning and testing) in your function `GetGitBranchName` after `#else // !USELIBGIT2` and new feature will be work in more wide situations.

ujos:
@akruphi Ok, please check now

akruphi:
> @akruphi Ok, please check now

Thank you. Very elegant compromise implementation.

Now we wait @elfmz decision about all.

If possible, please also add description about feature in help in section `@CommandPrompt` in alphabetical order in  https://github.com/elfmz/far2l/blob/1c91cfa5362fe2a62446ac6ae10018508bb6ff90/far2l/bootstrap/scripts/FarEng.hlf.m4#L2790-L2800 https://github.com/elfmz/far2l/blob/1c91cfa5362fe2a62446ac6ae10018508bb6ff90/far2l/bootstrap/scripts/FarHun.hlf.m4#L2436 https://github.com/elfmz/far2l/blob/1c91cfa5362fe2a62446ac6ae10018508bb6ff90/far2l/bootstrap/scripts/FarRus.hlf.m4#L2847 https://github.com/elfmz/far2l/blob/1c91cfa5362fe2a62446ac6ae10018508bb6ff90/far2l/bootstrap/scripts/FarUkr.hlf.m4#L2682

far2l/src/mix/GitTools.cpp
```diff
@@ -0,0 +1,45 @@
+#include "headers.hpp"
+#include "GitTools.hpp"
+
+#if defined(USELIBGIT2)
+#	include <git2.hpp>
+
+static git2::environment GitEnvironment;
+#else
+#	include <sys/stat.h>
```
akruphi, line 9: may be not need

ujos:
I think I'm done

far2l/CMakeLists.txt
```diff
@@ -225,6 +226,15 @@ else()
...
 endif()
 
+find_package(Libgit2)
+if (TARGET libgit2)
+    message(STATUS "Building with libgit2")
+    target_compile_definitions(far2l PRIVATE -DUSELIBGIT2)
+    target_link_libraries(far2l
+        PRIVATE libgit2::libgit2
+    )
+endif()
```
akruphi, line 236: In my opinion (it may be wrong) better try use libgit2 only if it need and check explicit flag `USELIBGIT2`:
```cmake
if (NOT USELIBGIT2)
    message(STATUS "Building without try check and use libgit2 due to USELIBGIT2=${USELIBGIT2}")
else()
    find_package(Libgit2) and
 and all you code here
endif()
```

ujos: My approach is to quietly find libgit2. if succeeded, then use it, otherwise fallback to the `git` command line invocation. 

Your way is more explicit. I would modify it a bit:

```cmake
if (NOT USELIBGIT2)
    message(STATUS "Building without try check and use libgit2 due to USELIBGIT2=${USELIBGIT2}")
else()
    find_package(Libgit2 REQUIRED)   # Require ligit2 now because of USELIBGIT2
   # and all you code here
endif()
```

akruphi: As you wish. :smile_cat:

ujos:
Are there any issues with this PR? I can decrease the scope and leave only the code which invokes git command line tool

ujos:
The latest commit solves two issues which probably won't happen on the GCC 11.5+. I'm not sure what is the target compiler though.

elfmz:
Still unclear why using extra lib? lets continue simpliest solution (with popen) and add complications once needed
Also can unify with this code: https://github.com/elfmz/far2l/blob/master/far2l/src/panels/infolist.cpp#L799

PR #2739, opened 2025-05-07
Fix cmake warnings

ctapmex:
errors appear on the latest version of cmake, such as:
```
CMake Warning (dev) at far2l/bootstrap/CMakeLists.txt:35 (add_custom_command):
  COMMENT requires exactly one argument, but multiple values or COMMENT
  keywords have been given.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
```

```
CMake Warning (dev) at far2l/bootstrap/CMakeLists.txt:55 (add_custom_command):
  The following keywords are not supported when using
  add_custom_command(TARGET): DEPENDS.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
```
Using 'DEPENDS', you need to specify the dependencies that should be generated by other steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
+include(${CMAKE_SOURCE_DIR}/cmake/FuncCopyResources.cmake)
 
 cmake_minimum_required (VERSION 3.5.0)
 
--- a/NetRocks/CMakeLists.txt
+++ b/NetRocks/CMakeLists.txt
@@ -102,10 +102,18 @@ set_target_properties(NetRocks-SHELL
     SUFFIX ".broker")
 target_compile_options(NetRocks-SHELL PUBLIC -DNETROCKS_PROTOCOL)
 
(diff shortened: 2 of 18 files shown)


ctapmex:
`depends` works or can be specified only
for `add_custom_command(OUTPUT..` . Only for them the dependencies are checked. For `add_custom_command(TARGET ...` the command is being unconditionally executed , as I understand it.
it should have worked like this before, but it didn't output errors.

https://cmake.org/cmake/help/latest/command/add_custom_command.html

shmuz:
A few months ago I deleted those DEPENDS on far2m but later discovered that some plugin was not updated properly after rebuild. I did not investigate further, just put the DEPENDS back.

ctapmex:
okay, now we are monitoring the changes of all auxiliary files (resources). And we copy them when they are changed. This is exactly the way it works, and it doesn't contradict the cmake syntax.

ctapmex:
It's not all good yet, I'm looking

ctapmex:
done

PR #2742, opened 2025-05-08
Improvements of kitty protocol implementation

unxed:
Our implementation of the Kitty protocol is still not fully compliant with the specification. One of the discrepancies was that when the Shift key was pressed, the `unicode-key-code` field contained values that didn't conform to the spec for all keys except alphabetic ones. After this commit, that field will contain 0 in such cases — which is not ideal, but still better than providing incorrect values. In any case, for character codes, applications should rely on the `shifted key` or `text-as-codepoints` fields that provide chars taking in mind Shift state, so this change should not affect real-world application behavior (tested it in the kitty mode of far2l and also in turbo).

### DIFF
--- a/far2l/src/vt/vtshell_translation_kitty.cpp
+++ b/far2l/src/vt/vtshell_translation_kitty.cpp
@@ -48,11 +48,11 @@ https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
 std::string VT_TranslateKeyToKitty(const KEY_EVENT_RECORD &KeyEvent, int flags, unsigned char keypad)
 {
 	std::string out;
-	int shifted = 0;
-	int modifiers = 0;
+	unsigned int shifted = 0;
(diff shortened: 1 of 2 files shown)


far2l/src/vt/vtshell_translation_kitty.cpp
```diff
@@ -222,15 +222,53 @@ std::string VT_TranslateKeyToKitty(const KEY_EVENT_RECORD &KeyEvent, int flags,
...
 		}
 	}
 
+	switch (KeyEvent.wVirtualKeyCode) {
+
+		// top row
+		case VK_OEM_3:      base = '`'; break;
+		// ...digits...
+		case VK_OEM_MINUS:  base = '-'; break;
+		case VK_OEM_PLUS:   base = '-'; break;
```
akruphi, line 251: May be here '+' instead '-'?

PR #2746, opened 2025-05-10
From far2m

akruphi:
FAR API: fix crashing on invalid dialog coordinates (from far2m & Far3)
 from: https://github.com/shmuz/far2m/commit/9868000e6a96ddee05fa0623240484865369ae76
 & https://github.com/FarGroup/FarManager/blob/5f71f81412163821955049811aba544be5cbdcae/far/plugapi.cpp#L947-L951

### DIFF
--- a/far2l/src/plug/plugapi.cpp
+++ b/far2l/src/plug/plugapi.cpp
@@ -1003,9 +1003,10 @@ static HANDLE FarDialogInitSynched(INT_PTR PluginNumber, int X1, int Y1, int X2,
 		return hDlg;
 
 	// ФИЧА! нельзя указывать отрицательные X2 и Y2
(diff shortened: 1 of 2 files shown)


far2l/src/plug/plugapi.cpp
```diff
@@ -1003,7 +1003,11 @@ static HANDLE FarDialogInitSynched(INT_PTR PluginNumber, int X1, int Y1, int X2,
 		return hDlg;
 
 	// ФИЧА! нельзя указывать отрицательные X2 и Y2
-	if (X2 < 0 || Y2 < 0)
+	const auto fixCoord = [](int first, int second) { return (first < 0 && second == 0) ? 1 : second; };
```
elfmz, line 1006: is this lambda really needed? I think just using same expression in place will be more readable

PR #2750, opened 2025-05-14
Align implementation to recent spec author recomendations

unxed:
Switched back to using shifted key value if unshifted is undetectable, as recommented by spec author in
https://github.com/kovidgoyal/kitty/issues/8620#issuecomment-2878530117


elfmz:
эм, а изменения в коде предполагаются в этом ПР?)

PR #2754, opened 2025-05-21
Colorer plugin refactoring

ctapmex:
Colorer plugin refactoring


colorer/src/pcolorer2/ChooseTypeMenu.cpp
```diff
@@ -127,8 +126,7 @@ void ChooseTypeMenu::MoveToFavorites(size_t index)
 size_t ChooseTypeMenu::AddFavorite(const FileType* fType)
 {
   size_t i;
-  for (i = favorite_idx; i < ItemCount && !(Item[i].Flags & MIF_SEPARATOR); i++)
```
elfmz, line 130: suggest to move out this 3 times used loop into separate function (LookupNextNonSeparator() ?)

PR #2760, opened 2025-05-25
readme & changelog actualization

akruphi:


akruphi:
and add help info after discussion in https://github.com/elfmz/far2l/issues/2761

PR #2797, opened 2025-07-08
Update trash.sh

tempergate:
added trash rs and trash-cli py support
https://github.com/oberblastmeister/trashy/tree/master  added trashy(all distros) and trash(the same for NixOS)
https://github.com/andreafrancia/trash-cli

### DIFF
--- a/far2l/bootstrap/trash.sh
+++ b/far2l/bootstrap/trash.sh
@@ -17,10 +17,10 @@ fi
 if command -v kioclient >/dev/null 2>&1; then
 	kioclient move "$1" trash:/ 2>"$2"
 
-elif command -v trash put "$1" 2>"$2"; then
+elif command -v trash >/dev/null 2>&1; then
 	trash put "$1" 2>"$2"
 
-elif command -v trashу put "$1" 2>"$2"; then
+elif command -v trashу >/dev/null 2>&1; then
 	trashу put "$1" 2>"$2" 
 
 elif command -v gio >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


far2l/bootstrap/trash.sh
```diff
@@ -14,6 +14,12 @@ if [ -x ~/.config/far2l/trash.sh ]; then
 . ~/.config/far2l/trash.sh
 fi
 
+if command -v trash put "$1" 2>"$2"; then
```
mitya57: There is no need to pass arguments to `command -v trash`, they will be ignored anyway. Also let's ignore stdout and redirect stderr to stdout, for consistency with the blocks below.

```suggestion
if command -v trash >/dev/null 2>&1; then
```

far2l/bootstrap/trash.sh
```diff
@@ -14,6 +14,12 @@ if [ -x ~/.config/far2l/trash.sh ]; then
...
+	trash put "$1" 2>"$2"
+
+if command -v trashу put "$1" 2>"$2"; then
```
mitya57: Similarly here, and also `s/if/elif/` (otherwise the previous `if` is not closed).

```suggestion
elif command -v trashу >/dev/null 2>&1; then
```

far2l/bootstrap/trash.sh
```diff
@@ -14,6 +14,12 @@ if [ -x ~/.config/far2l/trash.sh ]; then
...
+	trashу put "$1" 2>"$2"
+
 if command -v kioclient >/dev/null 2>&1; then
```
mitya57, line 23: This needs to be changed to `elif` as well.
```suggestion
elif command -v kioclient >/dev/null 2>&1; then
```

mitya57 (changes requested):

PR #2819, opened 2025-07-26
fix black.hrd for non-higlighted files

tempergate:
When 'CrossDraw=1' in non highlighted files, text color was #000000 in front of #313b30. It is unreadable.  Changed text color from  #000000 to #fbc9ff to make it readble


ctapmex:
from me - ok. accepted into the main repository

tempergate:
closed as it added in  https://github.com/elfmz/far2l/pull/2827

PR #2823, opened 2025-07-27
fix perl colorize error

ctapmex:
fix perl colorize error (description https://github.com/colorer/Colorer-schemes/issues/187  ) after c7e1b8b5bb15784aca678e0d9795130fcbac51f7


ctapmex:
include to https://github.com/elfmz/far2l/pull/2827

PR #2859, opened 2025-09-03
Fix tab navigation over radio buttons

exkrexpexfex:
see e.g. F9->Options->Interface settings where date format couldn't be changed when navigating with tab


exkrexpexfex:
Current behavior is standard according to [W3C]( https://www.w3.org/wiki/RadioButton), no need to change

PR #2861, opened 2025-09-04
Copy overwrite dialog minor

akruphi:
Idea from https://github.com/elfmz/far2l/discussions/2860


akruphi:
Closed because #2862 is more common

PR #2871, opened 2025-09-17
Adjust colors for big-endian

### LINKED ISSUE #2870: Big-endian build: wrong colors, no text input with --notty
Looks like colors are wrong on Big-endian?

Image

Also with `--notty` keyboard input does not work (could be that wx build needs some adjustment, at least the executable does not link to wx or x11).

exkrexpexfex:
fix #2870

### DIFF
--- a/far2l/src/color.hpp
+++ b/far2l/src/color.hpp
@@ -32,30 +32,53 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
-typedef struct rgbcolor_s
-{
-	union {
-		struct {
-			uint8_t r;
-			uint8_t g;
-			uint8_t b;
-			uint8_t a;
-		};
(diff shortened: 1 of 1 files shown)


mitya57:
I can confirm that it fixes colors on a big endian system. Thank you!

barracuda156:
@exkrexpexfex Thank you! I will try rebuilding.

barracuda156 (approved):
I can also confirm this fixes colors on macOS on powerpc. Thank you!

akruphi:
This fix solve only base problem with 256-bit colors.

For RGB and alpha-colors we need add smart corrections for macros `RGB`, `GetRValue`, ... , `ATTR_RGBBACK_NEGF2` in **far2l/src/color.hpp**

For **big-endian** also in all code we must verify all bitwise and shift (`<<`, `>>`) operations if they apply for **more then 1 byte** (usually they do not endian-sensitive but may be corner cases).
Quick search for ` >> ` found such operators in **157 files**: base far (`far2l/src/pick_colorRGB.cpp`, `far2l/src/options.cpp`, `far2l/src/hilight.cpp`...), WinPort, colorer, etc...

PR #2878, opened 2025-09-22
hexitor & plugins api

m32:
- less items to port from far3 version of hexitor
- expose color dialog for plugins from far2l API

### DIFF
--- a/far2l/far2sdk/farplug-wide.h
+++ b/far2l/far2sdk/farplug-wide.h
@@ -2235,7 +2235,6 @@ struct PluginStartupInfo
 	FARAPIADVCONTROL       AdvControl;
 	FARAPIADVCONTROL       AdvControlAsync;
 	FARAPIINPUTBOX         InputBox;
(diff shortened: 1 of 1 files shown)


far2l/far2sdk/farplug-wide.h
```diff
@@ -2230,6 +2235,7 @@ struct PluginStartupInfo
 	FARAPIADVCONTROL       AdvControl;
 	FARAPIADVCONTROL       AdvControlAsync;
 	FARAPIINPUTBOX         InputBox;
+	FARAPICOLORDIALOG      ColorDialog;
```
elfmz: better move it to the end so at least be back compatible whenever possible

m32: I didn't write it myself, but I see the point of the place - it's in the same place as in farplug-mb.h, after InputBox.
Then the FarStandardFunctions structure/api is preserved.

m32: suggested change implemented

PR #2922, opened 2025-09-28
trying to fix crash in DumpToFile

unxed:
see #2921


unxed:
fixed in https://github.com/elfmz/far2l/commit/da62590fddae1d2f6e4f49cd4a1e703de5f59bd6

PR #2926, opened 2025-09-29
Make arclite plugin compatible with uClibc-ng

Mi-Walkie-Talkie-by-Darkhorse:
Do not use the 'dlinfo' function when building arclite plugin with uClibc-ng, as this function is not present in the library.

### DIFF
--- a/arclite/src/Patch7zCP.cpp
+++ b/arclite/src/Patch7zCP.cpp
@@ -778,7 +778,7 @@ bool get_faddrs(void *handle)
 
 static bool patch_plt(void *handle)
 {
-#if defined(__APPLE__)
+#if defined(__APPLE__) || defined(__UCLIBC__)
 	return false;
 #else
 	struct link_map *map;


elfmz:
oh, but why it doing such sort of things at all? (i mean functions hooking). At least it will not work on certain systems (W^X protection or systems with hard-readonly code segments)

Mi-Walkie-Talkie-by-Darkhorse:
> oh, but why it doing such sort of things at all? (i mean functions hooking). At least it will not work on certain systems (W^X protection or systems with hard-readonly code segments)

It looks like it _ actually _ doesn't do anything - the only 'patch_plt' function call is commented out.
Unfortunately, this unreachable function still produces compile error with uClibc :(

elfmz:
but there is patch_addr that is doing that

Mi-Walkie-Talkie-by-Darkhorse:
> but there is patch_addr that is doing that

Да, проглядел, действительно выглядит кошмарно :( 
Однако, не я это писал, не мне и выкидывать...

Issue #2950, opened 2025-10-03
[hexitor] Crashes and non-working style setup


tempergate:
<s>1) mouse1 on exit element leads to crash</s>
<img width="2560" height="1600" alt="Image" src="https://github.com/user-attachments/assets/0e9eef59-6b72-489c-ba4b-921ce1569b7d" />

[far-log.txt](https://github.com/user-attachments/files/22683780/far-log.txt)

[crash-log.txt](https://github.com/user-attachments/files/22683781/crash-log.txt)

<s>2) After the last update far startup significantly slows down (probably not related to hexitor)</s>

3) non working style backgrounds setups in hexitor settings. It resets variables after closing colors selector
<img width="1520" height="683" alt="Image" src="https://github.com/user-attachments/assets/5d480165-df4e-4f85-b5cd-91480ecec6bd" />

[config-ini.txt](https://github.com/user-attachments/files/22683873/config-ini.txt)

@m32

m32:
I changed the dialog using colors, but I don't know if that's what you wanted. It behaves as I imagine it should.
offset addresses - address within the file - color changes
editable data - color of the current cell - color changes
updated data - marks places where data has changed - color changes
 fix submited

tempergate:
> I changed the dialog using colors

<s>If you select RGB in background it wont set the color. After you close the color picker dialog, it dismisses the RGB select and wipes HEX parameter from line edit</s>

3. colors change works  https://github.com/elfmz/far2l/pull/2951

tempergate:
1) Crash fixed in https://github.com/elfmz/far2l/pull/2951

tempergate:
3. stile settings are saving, but resetting after far process ending. 
Need additional testing. 
But current issue is closing, as it was solved

PR #2951, opened 2025-10-03
hexitor

m32:
- exit with mouse generates crash
- preserve colors when cancel on whole dialog was chosen


tempergate:
fixes https://github.com/elfmz/far2l/issues/2950
need to be merged to prevent crash

Issue #3004, opened 2025-10-21
[NetRocks] [SMB] SMBv1 support expected from server


russiandesman:
SMBv1 is deprecated and recommended for removal from the list of supported protocols on servers.
However, when actually removed -- Netrocks stops to connect to such servers.
smbclient and mount.cifs are working fine with both enabled and disabled v1 support on server negotiating the available protocol with server.

The reason is probably smbc_init() in ProtocolSMB::ProtocolSMB
The proposal is to pass v1 support requirement via protocol options and use smbc_new_context() if v1 support is not required

russiandesman:
SMB protocol logging added in #3008
After investigation it appears that moving to "modern" API is not strictly necessary, although I made a prototype with checkbox in [protocol options], let me know if you want to consider its adoption

The rootcause was that I mistyped the password, so there is no issue with current implementation

elfmz:
so, closing, if no issue

PR #3008, opened 2025-10-23
Add logging to SMB protocol implementation

russiandesman:
SMB protocol logging added to ease troubleshooting
related to https://github.com/elfmz/far2l/issues/3004

### DIFF
--- a/NetRocks/src/Protocol/SMB/ProtocolSMB.cpp
+++ b/NetRocks/src/Protocol/SMB/ProtocolSMB.cpp
@@ -235,7 +235,7 @@ void ProtocolSMB::Rename(const std::string &path_old, const std::string &path_ne
 
 void ProtocolSMB::SetTimes(const std::string &path, const timespec &access_time, const timespec &modification_time)
 {
-	NR_DBG("path: %s, atime: %ld, mtime: %ld", path.c_str(), access_time.tv_sec, modification_time.tv_sec);
+	NR_DBG("path: %s, atime: %jd, mtime: %jd", path.c_str(), (intmax_t)(access_time.tv_sec), (intmax_t)(modification_time.tv_sec));
 	struct timeval times[2] = {};
 	times[0].tv_sec = access_time.tv_sec;
 	times[0].tv_usec = suseconds_t(access_time.tv_nsec / 1000);


NetRocks/src/Erroring.h
```diff
@@ -2,10 +2,28 @@
...
 #include <stdexcept>
 #include <string>
+#include <cstdarg>
+#include <cstdio>
+#include <ctime>
 
 
 extern unsigned char g_netrocks_verbosity;
 
+void NetrocksLog(const int threshold, const char* file, const int line, const char* caller, const char* fmt, ...);
```
exkrexpexfex, line 13: adding `__attribute__((format(printf, 5, 6)))` here will help to reveal format string issues to be fixed

NetRocks/src/Erroring.cpp
```diff
@@ -20,6 +21,22 @@ unsigned char InitNetrocksVerbosity()
 
 unsigned char g_netrocks_verbosity = InitNetrocksVerbosity();
 
+void NetrocksLog(const int threshold, const char* file, const int line, const char* caller, const char* fmt, ...)
+{
+	if (g_netrocks_verbosity > threshold)
+	{
+		std::time_t now = std::time(nullptr);
+		std::tm* timeinfo = std::localtime(&now);
+		char timestamp[20];
```
pixeldexter: [nit] instead of counting characters manually, you could do this
```suggestion
		char timestamp[sizeof("YYYY-mm-dd HH:MM:SS")];
```
Takes info account terminating \0 as well

NetRocks/src/Erroring.h
russiandesman: thanks, done

NetRocks/src/Erroring.cpp
russiandesman: reworked to avoid this necessity

NetRocks/src/Protocol/SMB/ProtocolSMB.cpp
```diff
@@ -163,73 +178,98 @@ static int ProtocolSMB_GetInformationInternal(FileInformation &file_info, const
...
+	if (rc != 0) {
+		NR_ERR("smbc_rename failed, %d: %s", errno, strerror(errno));
 		throw ProtocolError("Rename error", errno);
+	}
 }
 
 
 void ProtocolSMB::SetTimes(const std::string &path, const timespec &access_time, const timespec &modification_time)
 {
+	NR_DBG("path: %s, atime: %ld, mtime: %ld", path.c_str(), access_time.tv_sec, modification_time.tv_sec);
```
akruphi: May be for **time_t** type use crossplatform
* `"%jd",  (intmax_t)access_time.tv_sec` (see https://en.cppreference.com/w/c/chrono/time_t )
* or `"%" PRIdMAX "text", (intmax_t)access_time.tv_sec` (it may be better, see https://en.cppreference.com/w/c/types/integer.html )

russiandesman: First syntax is ugly but ok, second one was born in hell definitely, so first one implemented

PR #3009, opened 2025-10-23
hexitor - minor: "Screen" label for F12 and help improvement

akruphi:


tempergate:
<s>**bug)** mouse1 on 12Screen element (key on keybar) not working
expected opening the same dialog as for f12 press</s> works with patch bellow

akruphi:
> mouse1 on 12Screen element (key on keybar) not working
> expected opening the same dialog as for f12 press

Now key **F12** processing via far2l manager, not inside hexitor.
Here we need advise from @m32 how correctly transfer from mouse action `KEY_F12` inside `bool editor::move_handle_mouse(` or inside `bool editor::sckey_handle(` to global manager.

m32:
PR pushed

tempergate:
> PR pushed

it works

PR #3010, opened 2025-10-23
added Note about use Wayland and OSC 52 under Wayland

tempergate:
It describes one little confusion for a user, that will start using terminal version under Wayland, as x|xi extensions are working not correctly under all Wayland composers and provides a solution to solve the problem


unxed:
Hmm, it seems like the clipboard used to work fine through Xwayland. Is it a specific composer issue? The ticket would

tempergate:
> Hmm, it seems like the clipboard used to work fine through Xwayland. Is it a specific composer issue? The ticket would

xwayland could be loaded in background by DEs like KDE or GNOME, but it is not guaranteed for every Window manager/Composer

unxed:
Isn't it time for #2040 @elfmz ? :)

tempergate:
reopened as it is actual problem for every `--tty` instance under `wayland`

PR #3011, opened 2025-10-24
hexitor: forward keys from dialog to framemanager

m32:
forward keys CTRL+TAB, CTRL+SHIFT+TAB, F12 from dialog to framemanager as required by PR #3009

### DIFF
--- a/hexitor/src/editor.cpp
+++ b/hexitor/src/editor.cpp
@@ -796,20 +796,8 @@ bool editor::sckey_handle(const int key)
 			}
 			break;
 
-		case KEY_TAB:
-			if (ctrl_pressed && !alt_pressed && shift_pressed) {
-				int KeyArray[] = {KEY_CTRLSHIFTTAB};
-				_PSI.SendDlgMessage(_dialog, DM_KEY, 1, (LONG_PTR)&KeyArray);
-				even_handled = true;
-			} else if (ctrl_pressed && !alt_pressed && !shift_pressed) {
-				int KeyArray[] = {KEY_CTRLTAB};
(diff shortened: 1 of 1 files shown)


tempergate:
works correctly

hexitor/src/editor.cpp
```diff
@@ -796,6 +796,26 @@ bool editor::sckey_handle(const int key)
...
+			if (ctrl_pressed && !alt_pressed && shift_pressed) {
+				int KeyArray[] = {KEY_CTRLSHIFTTAB};
+				_PSI.SendDlgMessage(_dialog, DM_KEY, 1, (LONG_PTR)&KeyArray);
+				even_handled = true;
+			} else if (ctrl_pressed && !alt_pressed && !shift_pressed) {
+				int KeyArray[] = {KEY_CTRLTAB};
+				_PSI.SendDlgMessage(_dialog, DM_KEY, 1, (LONG_PTR)&KeyArray);
+				even_handled = true;
+			}
+			break;
```
akruphi: Why `KEY_CTRLTAB` and `KEY_CTRLSHIFTTAB` add here and in far2l/src/dialog.cpp? Before this PR these keys processing properly via far2l manager.

I think (may be wrong) only additional processing `KEY_F12` need here.

tempergate: > I think (may be wrong) only additional processing `KEY_F12` need here.

maybe it is for binds like Control|Alt|Shift + F*

m32: @akruphi onyl code for F12 is needed - fixed

PR #3016, opened 2025-10-26
Update far2lOverlays.nix to fix 7z version and changed actual build

tempergate:
fixed 7z version back
updated far2l commit and md5 for actual version 2.7.0


elfmz:
in future need to make this automated from version_bump_patch.sh but for now need to release this somehow 🫠

PR #3017, opened 2025-10-26
hexitor: prevent far2l crash on window resize while hexitor is open

m32:


elfmz:
i already worked around this here: https://github.com/elfmz/far2l/commit/b4f641c8c99c62e37e5505302ddc8364b132bdd8
Is this still neccessary? Or better than mine fix (then undo it)

akruphi:
> i already worked around this here: [b4f641c](https://github.com/elfmz/far2l/commit/b4f641c8c99c62e37e5505302ddc8364b132bdd8) Is this still neccessary? Or better than mine fix (then undo it)

The morning commit fixed only crash, but now inside openning hexitor resize to small width leads to undesirable line wraps. This PR fix such behavior.

m32:
@elfmz hexitor

hexitor has limit for labels = 6 chars, your fix works probably only for this case
mine works with buffers allocated for status line, buttons line while resizing hexitor window

Issue #2379, opened 2024-09-08
NetRocks: SMB filename problems


faerot:
When pressing Ctrl+F on file in NetRocks, it inserts URL of this format:
```<smb:192.168.1.50>/music/``` - note the extra ```<``` ```>``` symbols which prevent usage of such url both in other programs and in far2l as well, because if you enter such url in far2l command line and press "enter", the url is not processed.
It should insert ```smb://192.168.1.50/music/``` instead. This type of url works in other programs and in far too.

Would like to also request supporting ```Shift+Enter``` and ```Context Menu``` button on SMB files.

faerot:
I have updated the old far2l from ubuntu repository to the one built from source and now saved folder bookmarks that point to SMB directories stopped working. It reports error "Not a directory (20)" when trying to use it. If have tried to manually edit saved ``<smb:192.168.1.50>/music/`` to the ones accepted in command line like ``smb://192.168.1.50/music/`` but it did not help.

PR #3021, opened 2025-10-27
Add ability to provide URLs from plugins. Use it in NetRocks

russiandesman:
Partially addresses https://github.com/elfmz/far2l/issues/2379

Pressing Ctrl+F now inserts an URL usable for both far2l and external tools

### DIFF
--- a/NetRocks/src/Host/Host.h
+++ b/NetRocks/src/Host/Host.h
@@ -17,7 +17,6 @@ struct IHost : IProtocol
 	};
 
 	virtual std::string SiteName() = 0; // MT-safe, human-readable site's name
-	virtual std::string SiteURL() = 0; // MT-safe, site's URL
 	virtual void GetIdentity(Identity &identity) = 0; // MT-safe, returns connection host identity details
 
 	virtual std::shared_ptr<IHost> Clone() = 0; // MT-safe, creates clone of this host that will init automatically with same creds
--- a/NetRocks/src/Host/HostLocal.cpp
+++ b/NetRocks/src/Host/HostLocal.cpp
@@ -55,16 +55,6 @@ std::string HostLocal::SiteName()
 #endif
 }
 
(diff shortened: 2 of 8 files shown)


akruphi:
Inside far2l **Ctrl+Enter** and **Ctrl+F** put path with **escaping** into command line for fully use inside far2l.

For copy to external I suggest use **Ctrl+Ins** and **Ctrl+Alt+Ins** to direct copy path to clipboard without escaping.

elfmz:
1 there is no need to add IHost::GetURL - just use existing IHost::GetIdentity from PluginImpl::UpdatePathInfo()
2 better replace OpenPluginInfo::Reserved field with CurURL, or put CurURL just before Reserved - this will provide more compatibility (even at source code level)

russiandesman:
> 1 there is no need to add IHost::GetURL - just use existing IHost::GetIdentity from PluginImpl::UpdatePathInfo() 2 better replace OpenPluginInfo::Reserved field with CurURL, or put CurURL just before Reserved - this will provide more compatibility (even at source code level)

Done

russiandesman:
please note that there is still a bug -- both CurDir and CurURL are relative to working dir provided by user explicitly or implicitly.
The most straightforward example is ssh session. With empty working dir we start from user's home (which is not guaranteed to be at /home/username, it's frankly arbitrary) and then we walk here and there, and can walk upper...
Simple method to retrieve PWD in plugin would be nice to have...

russiandesman:
> Inside far2l **Ctrl+Enter** and **Ctrl+F** put path with **escaping** into command line for fully use inside far2l.
> 
> For copy to external I suggest use **Ctrl+Ins** and **Ctrl+Alt+Ins** to direct copy path to clipboard without escaping.

IMO the whole point for Ctrl+F lays in its' mnemonic -- **F**ull, and full to me means "usable externally".

PR #3022, opened 2025-10-28
Run in editor mode when called as /usr/bin/editor symlink

mitya57:
In Debian, `/usr/bin/editor` is managed by [alternatives system][1], and I am going to register far2ledit as a possible alternative for that.

With `editor` → `far2ledit` → `far2l` symlinks chain, far2l does not start in editor mode. This change helps with that.

Note: Debian's mc has similar logic, but via [a patch][2] (upstream refused to accept this change).

[1]: https://wiki.debian.org/DebianAlternatives
[2]: https://salsa.debian.org/debian/mc/-/blob/master/debian/patches/alt_editor.patch

### DIFF
--- a/far2l/src/main.cpp
+++ b/far2l/src/main.cpp
@@ -87,7 +87,7 @@ int DirectRT = 0;
 
 static void print_help(const char *self)
 {
-	bool is_far2ledit = strstr(self, "far2ledit") != NULL;
+	bool is_far2ledit = strstr(self, "edit") != NULL;
 	printf("FAR2L Version: %s\n"
 			"FAR2L - two-panel file manager, with built-in terminal and other usefulness'es\n"
 			"Usage: %s [switches] [-cd apath [-cd ppath]]\n"
@@ -457,7 +457,10 @@ int FarAppMain(int argc, char **argv)
 	}
 
 	// run by symlink in editor mode
-	bool is_far2ledit = strstr(argv[0], "far2ledit") != NULL;
+	// The name can be "far2ledit" or "editor" (when far2ledit is chosen
+	// as the default editor in Debian-based systems).
+	// See https://github.com/elfmz/far2l/pull/3022.
+	bool is_far2ledit = strstr(argv[0], "edit") != NULL;
 	if (is_far2ledit) {
 		Opt.OnlyEditorViewerUsed = Options::ONLY_EDITOR;
 		if (argc > 1) {


far2l/src/main.cpp
```diff
@@ -457,7 +457,7 @@ int FarAppMain(int argc, char **argv)
 	}
 
 	// run by symlink in editor mode
-	bool is_far2ledit = strstr(argv[0], "far2ledit") != NULL;
+	bool is_far2ledit = strstr(argv[0], "edit") != NULL;
```
akruphi, line 463: May be for future add here explicit comment:
```cpp
// all "*edit*" process as editor, need for "far2ledit" and debian alternative "editor"
// (see https://github.com/elfmz/far2l/pull/3022 )
```

mitya57: Done.

far2l/src/main.cpp
```diff
@@ -87,7 +87,7 @@ int DirectRT = 0;
 
 static void print_help(const char *self)
 {
-	bool is_far2ledit = strstr(self, "far2ledit") != NULL;
+	bool is_far2ledit = strstr(self, "edit") != NULL;
```
elfmz, line 90: `strstr(PointToName(self), "edit") != NULL` will be safer

far2l/src/main.cpp
```diff
@@ -457,7 +457,10 @@ int FarAppMain(int argc, char **argv)
 	}
 
 	// run by symlink in editor mode
-	bool is_far2ledit = strstr(argv[0], "far2ledit") != NULL;
+	// The name can be "far2ledit" or "editor" (when far2ledit is chosen
+	// as the default editor in Debian-based systems).
+	// See https://github.com/elfmz/far2l/pull/3022.
+	bool is_far2ledit = strstr(argv[0], "edit") != NULL;
```
elfmz, line 463: see above note on PointToName()

far2l/src/main.cpp
mitya57: I can change, but I think we already strip slashes this way:
```c
    char *name = strrchr(argv[0], GOOD_SLASH);
    if (name)
        ++name;
    else
        name = argv[0];
```

elfmz: here you right, but in another case below still better to use PointToName

far2l/src/main.cpp
mitya57: Does not work as is:
```
/home/dmitry/packages/far2l/far2l/src/main.cpp: In function ‘int FarAppMain(int, char**)’:
/home/dmitry/packages/far2l/far2l/src/main.cpp:463:54: error: cannot bind non-const lvalue reference of type ‘FARString&’ to an rvalue of type ‘FARString’
  463 |         bool is_far2ledit = strstr(PointToName(argv[0]), "edit") != NULL;
      |                                                ~~~~~~^
In file included from /home/dmitry/packages/far2l/far2l/src/headers.hpp:161,
                 from /home/dmitry/packages/far2l/far2l/src/main.cpp:34:
/home/dmitry/packages/far2l/far2l/src/base/FARString.hpp:132:16: note:   after user-defined conversion: ‘FARString::FARString(const char*, UINT)’
  132 |         inline FARString(const char *lpszData, UINT CodePage=CP_UTF8) { Init(); Copy(lpszData, CodePage); }
      |                ^~~~~~~~~
In file included from /home/dmitry/packages/far2l/far2l/src/main.cpp:58:
/home/dmitry/packages/far2l/far2l/src/mix/pathmix.hpp:81:39: note: initializing argument 1 of ‘const wchar_t* PointToName(FARString&)’
   81 | const wchar_t *PointToName(FARString &strPath);
      |                            ~~~~~~~~~~~^~~~~~~
```

Probably because `PointToName` accepts either `FARString &` or `const wchar_t *`, but here we have `char *`. And it returns `wchar_t *` too, which is not suitable for `strstr`.

Maybe I can copy the logic of `char *name` that I mentioned above from `main()` to `FarAppMain()`?

It doesn’t look like there is an easy way to _pass_ it from one function to the other one.

PR #3024, opened 2025-10-28
[WIP] Tree panel amendments

exkrexpexfex:
* limited recursive scanning depth
* right arrow expands collapsed subtrees
* left arrow collapses expanded subtree
* ctrl+number expands all branches to the chosen depth
* automated memory management for ListData
* faster sorting for the default case when numeric sort is disabled


changelog.md
```diff
@@ -5,6 +5,9 @@ Only significant user-side changes are listed here
 or via `git log --no-merges --pretty=format:"%as: %B"`).
 
 ## Master (current development)
+* Tree panel: Option to exclude subtrees from scanning using a mask (default: hidden folders `.*`).
+  Option to set the maximum recursive scanning depth (default: 4). 
+  Right Arrow expands excluded subtrees, and Left Arrow now navigates one level up.
```
akruphi, line 10: "`* Tree panel: `" - in each line please add prefix which explain object of changes. Because in Debian/Ubuntu package text from md-changelog converted to plain text (during metainfo generation) and each new line with any indent need explanation.

exkrexpexfex: Shall I collapse it to a single line?

akruphi: > Shall I collapse it to a single line?

I have not opinion about it. As you wish. May be @mitya57 give advise.

mitya57: It's interpreted as Markdown, not as plain text. Like you wrote it currently has the same effect as if it were collapsed into a single line. The result is the following:
```html
        <ul>
          <li>Tree panel: Option to exclude subtrees from scanning using a mask (default: hidden folders <code>.*</code>).
  Option to set the maximum recursive scanning depth (default: 4).
  Right Arrow expands excluded subtrees, and Left Arrow now navigates one level up.</li>
        </ul>
```

A bit of self-advertisement: you can use [ReText](https://github.com/retext-project/retext) to check how it looks rendered, since it uses the same library (Python-Markdown) under the hood as `generate_metainfo.py`.

Screenshot

elfmz:
was there true need for MaxDepth ? there is already mechanisms to avoid infinite recursion due to symlinks etc
ScanTree is used also in many other places, like copying files, where such hard limit can cause data loss

exkrexpexfex:
> was there true need for MaxDepth ? there is already mechanisms to avoid infinite recursion due to symlinks etc ScanTree is used also in many other places, like copying files, where such hard limit can cause data loss

yes, removed hard limit 4096 not to affect other places. Trying to revive the folders tree functionality to be usable with large deep file structures, not sure it can be achieved without touching ScanTree at all

PR #3032, opened 2025-11-01
Fix missed dot-slash prefix for Ctrl+Enter

russiandesman:
fix regression after #3021
    Fix missed dot-slash prefix for Ctrl+Enter
    Factor out URL fetching from plugins

### DIFF
--- a/far2l/src/cfg/ConfigOpt.cpp
+++ b/far2l/src/cfg/ConfigOpt.cpp
@@ -390,7 +390,6 @@ const ConfigOpt g_cfg_opts[] {
 	{true,  NSecPanel, "CaseSensitiveCompareSelect", &Opt.PanelCaseSensitiveCompareSelect, 1},
 	{true,  NSecPanel, "ReverseSort", &Opt.ReverseSort, 1},
 	{false, NSecPanel, "RightClickRule", &Opt.PanelRightClickRule, 2},
-	{false, NSecPanel, "CtrlFRule", &Opt.PanelCtrlFRule, 1},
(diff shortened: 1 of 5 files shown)


elfmz:
As i understand FolderUpperCase and FileLowerCase intentially ignored now in CreateFullPathName as in *nix this options are sort of broken by nature? If so, then may be then better to remove them completely (also from UI), instead of leaving them half working?

akruphi:
@elfmz this PR is need to fix regression with "`./`" for **Ctrl+Enter**.
Removing UI legacy is more complex problem and may be done accurately in future.

elfmz:
I mean that function CreateFullPathName previously was taking FileAttr argument that was used to translate filename to upper/lower case depending on this settings. In this PR this argument is no more, meaning that this settings won't work anymore for cases where function is used. But they still in use in other code. So its better to remove it completely than leave such half-broken feature that even doesn't applicable for Unix at all

russiandesman:
> I mean that function CreateFullPathName previously was taking FileAttr argument that was used to translate filename to upper/lower case depending on this settings. In this PR this argument is no more, meaning that this settings won't work anymore for cases where function is used. But they still in use in other code. So its better to remove it completely than leave such half-broken feature that even doesn't applicable for Unix at all

Regarding removal from UI. There are two independent settings. 1) "apply panel UI settings to ctrl-enter/ctrl-f
And 2) use uppercase for folders and lowercase for files in panel UI.

First one is carefully removed now, there is no half-broken things remaining 
Second one does not harm and can address someone's habits, so has been kept.

PR #3044, opened 2025-11-05
hexitor i18n/vs-code

m32:
Creating plugins with translations is tedious if you don't know the languages ​​used in the plugin (even a little), hence the idea of ​​using VS-code and the i18n-ally plugin, as well as machine translators.

Another problem is that the development environment (VS-code) doesn't show whether a given text has already been translated or what the translation looks like, and that's where i18n-ally comes in.

Another problem is the need to track whether a variable is assigned to text in the lng file and on which line it is located – it's very easy to make a mistake (I've already had problems with several languages, not to mention a dozen). Hence the idea of ​​automatically creating the i18nindex.h file and simultaneously checking if the translation exists in each language. The Python script i18n2lng.py in the hexitor directory solves this problem.
screen-en
screen-ru
translator


m32:
The program code still looks the same and is independent of the user interface (TUI) design; only the setting in vs-code changes (changing the display language).

The magic lies in the I18N macro and the i18n-ally settings.

elfmz:
have conflicts..

m32:
conflicts resolved

PR #3064, opened 2025-11-12
edsort - sort selected block in editor

m32:
Sometimes it is useful to sort rows in the editor, and even better when you can specify which column to sort from.


akruphi:
Minor advises:
* If has not any selection or if selection mode is not `BTYPE_STREAM` I recommend show information message with warning
* If selections of last line is not full edsort make sort taking into account all last line but after sorting stay original last line after sorted lines. I see code https://github.com/elfmz/far2l/blob/a0496190522d60755a0a39a0d394a1c4a2acc20f/edsort/src/edsort.cpp#L102-L111 but edsort stay last line after this correction.
* For English far2l setting edsort show by **F1** Polish help

And please add info in `changelog.md` and in `README.md`.

m32:
@akruphi patch submitted

PR #3065, opened 2025-11-12
edsort - fixes

m32:


akruphi:
Thank you.

May be change current selection up to full line only after `{ OK }` from user? It more user-friendly if he choose Cancel.

m32:
@akruphi PR submitted

PR #3067, opened 2025-11-12
edsort: additional dialogue

m32:
allowing to choose how to treat the last not fully selected line


akruphi:
Other corner case when selection not from begin of 1st line but whole last line don't give warning and show main dialog:

изображение

m32:
fix pushed

PR #3075, opened 2025-11-13
edsort: fix for another corner case

m32:
the selection on the first line may also need to be corrected


edsort/configs/plug/edsort_ru.lng
```diff
@@ -9,6 +9,7 @@
 "Сортировка работает только с выбором блока."
 "Edsort может сортировать только целые строки,"
 "но последняя строка выбрана лишь частично."
```
akruphi: here should be 1st="первая": `"но первая строка выбрана лишь частично."`

m32: I have no idea how it should be. I know some letters/words, but I don't remember the grammar - the translation is from Google.
Could you correct it in the ru.json file and then run python ./i18n2lng.py? Or make a correction in my far2l?
Or specify the line to replace with what content ?

m32: @akruphi I didn't run the i18n2lng.py after translation - that was the problem.

Issue #2976, opened 2025-10-10
Weird emoji symbol displaying in WX


spvkgn:
⚠️ U+26A0
Шрифт DejaVu Sans Mono 14 или Ubuntu Mono.
Файл [warn.txt](https://github.com/user-attachments/files/22840213/warn.txt)

<img width="266" height="347" alt="Image" src="https://github.com/user-attachments/assets/59f113ba-b062-4020-ae53-930f881fb9df" />

PR #3096, opened 2025-11-19
TTY:Trying to detect variation selector 16 support

exkrexpexfex:
touch #2976

image_2025-11-19-01

far2l / far2m

"файловые панели"

Gnome Terminal 3.38.3

FAR2M, version 2.7.0-2025-11-18-5e5b7de6-beta Linux x86_64

FAR2L, version 2.7.0-2025-11-19-f1ed9e29-beta Linux x86_64

### DIFF
--- a/WinPort/src/Backend/WinPortMain.cpp
+++ b/WinPort/src/Backend/WinPortMain.cpp
@@ -136,20 +136,42 @@ int detect_vs16_width() {
 	tcsetattr(STDIN_FILENO, TCSAFLUSH, &raw);
 
 	auto readPos = []() {
-		std::string buf;
+		int row = -1, col = -1;
 		char c;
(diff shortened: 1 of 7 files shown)


elfmz:
-  this needed only in case of tty backend used, so move this check to under if (arg_opts.tty) { and only if nodetect is not specified
- note that utils project is static library, so such global variable will be per-binary, you set it for far2l binary, but for other plugins for which it can be important will have own instance of this variable, so potential solution could be just use exported variable and declare its import in utils project
- after each cursor request - put also request for terminal status (as mostly supported) readPos must be with timeout and must bail out if terminal status reply obtained . See TTYNegotiateFar2l as example of such logic

exkrexpexfex:
@elfmz  please check now

exkrexpexfex:
@elfmz if any issue please let me know

PR #3106, opened 2025-11-23
Correct Message width calculation (from far2m)

akruphi:
from: https://github.com/shmuz/far2m/commit/687636e5c2a2af5dc921fa3dd6636732bc80ddd1 and https://github.com/shmuz/far2m/commit/5e5b7de60fa1c3f8e4cf83fdbfbdddefeee44f33

### DIFF
--- a/far2l/src/message.cpp
+++ b/far2l/src/message.cpp
@@ -135,13 +135,14 @@ static int ShowMessageSynched(DWORD Flags, int Buttons, const wchar_t *Title, co
 	}
 
 	for (MaxLength = BtnLength, I = 0; I < StrCount; I++) {
-		if (static_cast<DWORD>(Length = StrLength(Items[I])) > MaxLength)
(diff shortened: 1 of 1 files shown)


far2l/src/message.cpp
```diff
@@ -135,13 +135,14 @@ static int ShowMessageSynched(DWORD Flags, int Buttons, const wchar_t *Title, co
 	}
 
 	for (MaxLength = BtnLength, I = 0; I < StrCount; I++) {
-		if (static_cast<DWORD>(Length = StrLength(Items[I])) > MaxLength)
-			MaxLength = Length;
+		auto Len = StrCellsCount(Items[I], StrLength(Items[I]));
```
elfmz: StrZCellsCount is there exactly to dont call StrLength each time. Also its slightly faster.

akruphi: Done

PR #3111, opened 2025-11-24
Editor mouse select

elfmz:
Simplier approach. KISS principle to remember (band also was good:))

### DIFF
--- a/far2l/src/editor.cpp
+++ b/far2l/src/editor.cpp
@@ -2100,8 +2100,7 @@ int Editor::ProcessKey(FarKey Key)
 		}
 		case KEY_CTRLU: {
 			UnmarkMacroBlock();
-			UnmarkBlock();
-			Show();
+			UnmarkBlockAndShowIt();
 			return TRUE;
 		}
 		case KEY_CTRLC:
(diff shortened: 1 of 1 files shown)


yulian5:
This one is much better. However here is the list of problems I see after quick test:

1. vertical block selection still has some cursor flickering
2. in vertical block selection expands by a few symbols to the left from cursor when selecting from right to left
3. in vertical block selection cursor is on position a few symbols shifted to the right from selection when selecting from left to right 
4. vertical block is preserved when wrap mode enabled and and then disabled again

elfmz:
addressed vblock issues, didnt test with wrap mode enabled yet

elfmz:
with  with wrap mode enabled  vblock selection seems doesnt work at all - nor by mouse, nore by Alt+Cursor, I suspect its an intended behavior, right?

elfmz:
BTW there is proposal to use right click for vertical block selection, cuz Alt+mouse is problematic on many systems (on my it drags window by default). Any reasons not to do this?

yulian5:
> vertical block is preserved when wrap mode enabled and and then disabled again

@elfmz Vertical block selection is fine. But one issue is remained. How to reproduce:
1. Select vertical block
2. Switch to wrap-mode
3. Selection is gone, however block is still selected. You can copy and paste invisible block.
4. Switch back to non-wrap mode and you can see the selected block again.

It should be an easy fix.

> BTW there is proposal to use right click for vertical block selection, cuz Alt+mouse is problematic on many systems (on my it drags window by default). Any reasons not to do this?

I think it's a good idea.

unxed:
> I suspect its an intended behavior, right?

Yep. I couldn't come up with a use case for this or how it should work.

> Any reasons not to do this?

No one from my side.

yulian5:
@elfmz I created PR [#3115](https://github.com/elfmz/far2l/pull/3115) with fix for vertical block selection in wrap mode

yulian5:
@elfmz Good job with this mouse selection!

PR #3118, opened 2025-11-27
changelog and help actualization

akruphi:


yulian5:
Thanks, @akruphi. Somehow I lost the changes in the help files, they were present in the first PR.

PR #3123, opened 2025-11-28
Checking Mach-O signature on executables detection

### LINKED ISSUE #3119: Problems with a.out files
Hello!

I'm using v2.7.0 BETA version of far2I [`far2l-2.7.0-beta-MacOS-11.2-universal.dmg`] ((https://github.com/elfmz/far2l/releases/download/v_2.7.0/far2l-2.7.0-beta-MacOS-11.2-universal.dmg) as I know on macos Sonoma 14.1.1.

If you try to compile the file using g++ to default binary file (a.out), you cannot run it. 

Image

However we can compile to not default binary file

link to video: https://disk.yandex.ru/i/X5RYHvlTg2wuMQ

exkrexpexfex:
Fix #3119

### DIFF
--- a/far2l/src/execute.cpp
+++ b/far2l/src/execute.cpp
@@ -146,6 +146,24 @@ public:
 			if (r > 4 && buf[0] == 0x7f && buf[1] == 'E' && buf[2] == 'L' && buf[3] == 'F') {
 				fprintf(stderr, "ExecClassifier('%s') - ELF executable\n", cmd);
 				_executable = true;
+			// Mach-O thin
+			} else if (r > 4 &&
+			  (
+				!memcmp(buf, "\xCE\xFA\xED\xFE", 4) || // MH_MAGIC
+				!memcmp(buf, "\xCF\xFA\xED\xFE", 4) || // MH_MAGIC_64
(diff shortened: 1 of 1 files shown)


far2l/src/execute.cpp
```diff
@@ -146,6 +146,24 @@ class ExecClassifier
...
+				fprintf(stderr, "ExecClassifier('%s') - Mach-O executable\n", cmd);
+				_executable = true;
+			// Mach-O FAT
+			} else if (r > 4 &&
+			  (
+				!memcmp(buf, "\xCA\xFE\xBA\xBE", 4) || // FAT_MAGIC
+				!memcmp(buf, "\xBE\xBA\xFE\xCA", 4)    // FAT_CIGAM
+			  )) {
+				fprintf(stderr, "ExecClassifier('%s') - Mach-O universal binary\n", cmd);
+				_executable = true;
```
akruphi, line 166: May be this code put inside `if defined(__APPLE__) || (__MACH__)`?
Or could be other system that run run mach-O binaries?

exkrexpexfex: Thought about this too, but considered it overcomplicating, anyway the OS won't run things it can't

PR #3126, opened 2025-11-28
Bringing NumLock back to macOS

exkrexpexfex:


elfmz:
such logic will turn SCROLLLOCK_ON and CAPSLOCK_ON forever
need to off them from old state when they not pressed with &= ~CAPSLOCK_ON, but not only |= CAPSLOCK_ON when pressed

exkrexpexfex:
@elfmz for me caps lock modifier keeps working as expected, whether it's changed when far2l is in focus or not. However since I don't understand the whole idea here, I reduced the change impact to the special macOS case as it was previously.
Please check new version

PR #3127, opened 2025-11-30
Simplifying coloring when line numbers enabled

### LINKED ISSUE #3125: Line Numbers and coloring
When line numbers enabled the default coloring of leading tabs is broken

Image
and when Word Wrap mode is enabled employ lines are impacted as well

Image

exkrexpexfex:
Fix #3125

### DIFF
--- a/far2l/src/edit.cpp
+++ b/far2l/src/edit.cpp
@@ -2528,7 +2528,6 @@ void Edit::ApplyColor()
 		// Пропускаем элементы у которых начало больше конца
 		if (CurItem.StartPos > CurItem.EndPos)
 			continue;
-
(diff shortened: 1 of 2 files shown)


elfmz:
i think, @ctapmex should look at this

yulian5:
@elfmz , @exkrexpexfex 

I tested this code and didn't find any problems with it. It fixes "tabs" problem and doesn't break anything else. 

This is high-quality refactoring from "screen" coordinates to "logical" ones. I tried to fix logical coordinates for colors but applying them to everything significantly simplifies the code but changes the logic as well.

Even if in case any problems in this code, we need to fix them, this approach is much better (3rd iteration already!)

Great job, @exkrexpexfex!

PR #3150, opened 2025-12-13
Fix alt+F10 display artifacts

### LINKED ISSUE #3145: [Bug] Render error on exit tree dialog
After exiting Tree dialog, Alt+F10, then Escape, it rerenders only active panel. The second panel will be wasted with already closed tree dialog until you wont switch to the second panel

Image

rev 1ed16ae84a397401252190dd817fe444b5cf0b2c

exkrexpexfex:
Fix #3145


tempergate:
fix works

PR #3157, opened 2025-12-19
New feature: support of two clipboard buffers at once (ICCCM-compliant) grab mouse selection and paste from primary

viklequick:
Added ICCCM-compliant support for mouse selection and paste from primary buffer

Supports 

* X11 / Wayland with both wx and SDL backends;
* terminals with OSC52: OSC52 Write when available to place mouse selections to primary buffer
* terminals with OSC52 Read (when enabled at the terminal emulator side) to paste from primary buffer.

The selection works everywhere:

* Editor
* Dialogs
* Command line.

Both copy and paste operations can be disabled from Options -> Interface settings dialog.

image

with options


far2l/Include/plugin.hpp
akruphi, line 1: Зачем тут этот новый пустой файл?

unxed:
Can you please make it write to both clipboard and primary as browsers do if no clipboard options are specified? So I could remove conflicting logic from #3215

viklequick:
> Can you please make it write to both clipboard and primary as browsers do if no clipboard options are specified? So I could remove conflicting logic from #3215

Sorry, but it is wrong behavior: change both clipboards at once. The main idea of two clipboards in X11  that  it are living completely independently. 

And surely, browsers do not corrupt clipboards; both clipboards works independently and I'm using this feature every day. 

It seems you have certain clipboard management application active that "synchronizes" both primary and clipboard.

unxed:
No. I definitely have no special clipboard managers. I have a problem: text copied in wx backend can not be pasted to tty backend, because they use different clipboard types. If your approach fixes that in any other way, I will just remove that part from my PR

viklequick:
> No. I definitely have no special clipboard managers. I have a problem: text copied in wx backend can not be pasted to tty backend, because they use different clipboard types. If your approach fixes that in any other way, I will just remove that part from my PR

Could you explain a case how to reproduce it?

By default, terminal emulators use the primary for all mouse-related activities, and the clipboard itself is bounded to Shift+Ctrl+{C,X,V}. 

But what is being used by bracketed paste etc -- it seems it is terminal emulator specific. So, if you make me a hint how to reproduce, what's wrong and how it should be, I'll try to find less heavy solution rather than premature synchronization of clipboards.

Thank you.

unxed:
https://github.com/elfmz/far2l/issues/3214

1. Copy multi-line text to the clipboard in the graphical far2l.
2. Go to the tty far2l in GNOME Terminal (running in Xi mode).
4. Try pasting with Shift+Ins.
5. Something completely different is pasted.

And this was the case even before #3212.

I'm testing [on this text](https://raw.githubusercontent.com/unxed/ap/refs/heads/main/ap.md).

viklequick:
> #3214
> 2. Go to the tty far2l in GNOME Terminal (running in Xi mode).

As we're discovered, the Shift+Insert is the hardcoded keybinding to gnome terminal and it is processed by VTE internally, we even does not receive the keys. So, we need to solve the GNOME terminal issue outside of far2l by making PR to VTE.

It means, the aggressive hack that synchronizes both lkeyboards will not help if the clipboard owner will be not the far2l-wx.

viklequick:
I've fixed code a bit to not make so aggressive attempts to open/close clipboard when we have nothing selected. Hope this will work better.

viklequick:
By the way, the code change is very simple and belongs to editor.cpp and config (with two options and its' processing respectively). Just dozen of lines of code.

But the glue is huge: it needs several files to be touched, as it need a new function in clipboards API: way to choose primary or clipboard buffer. I've copied the same way as for other functions by registering `ChooseClipboard` and added stubs everywhere for all clipboard implementations.

The implementation for wx Clipboard Backend is trivial: use wx API to choose clipboard (one function, one line). 

I've added workaround for Wayland as well: due to strange limitations, wxGTK cuts access to primary buffer, despite the gtk itself provides full support to it for the Wayland. So, I've added a workaround for this case: if we have wayland and no X11, we're using gtk directly just to operate with primary buffer. Hope in future wxWidgets will remove this strange limitation and we'll able to remove this workaround.

Just an example how 20 lines-code patch needs 20 extra files to be modified ;-)

viklequick:
By the way, it is ready for merge

elfmz:
1 Why its only works for editor but now global setting that affects clipboard operations everywhere?
2 adding bit flag to SetConsoleTweaks would be simplier that adding new API, see CONSOLE_OSC52CLIP_SET as example

viklequick:
> 1 Why its only works for editor but now global setting that affects clipboard operations everywhere?

The idea is the dynamical handling for both clipboard buffers simultaneously, not by choosing one for all time. 

So, for all cases except mouse selection we should use the standard clipboard as before; and mouse selection uses its' own buffer.  In this case, we have two clipboards operational at one time: mouse uses "primary", keyboard uses "clipboard" and both co-exist without impact to each other.

And I am thinking to add support for both clipboards not for file editor only. I thought to add the same behavior for dialog controls, for viewer and for command line console. 

I've chosen file editor as a first part of improvement for two reasons: first, it is already have all necessary functions to add 2nd clipboard quickly; and at second, despite to command line and file viewer changes it won't break anything.

Adding support for the edit fields in dialogs is a little bit easy, we even have the mouse selection implemented; I have planned to make new PR to it once this one will be merged.

Unfortunately, file viewer needs more work as it have no mouse selection; it seems it is a subject for new PR bit later.

And the console window already have mouse selection but it is vertical block and I guess it should be reworked first. So, maybe it will be nice to make it as separate PR.. 

> 2 adding bit flag to SetConsoleTweaks would be simplier that adding new API, see CONSOLE_OSC52CLIP_SET as example

Yeah, thank you for this one. For config, it will be much simpler than making a new API. 

Not sure if it is good way for our case., with dynamic change of the clipboard buffer; the new API method asks to open primary buffer only when it is necessary and then close it back immediately after the operation.  Surely it is not the best way; the additional flag for OpenClipboard looks like the better option.  But once I have tried to use this way, the amount of changes was almost the same., so I've created a new API method to be sure I'll not touch anything in uncommon use cases.

elfmz:
I think its better to make it universally but not only editor.. This feature of separate clipboard is not very easy to use itself, but if it will be also only in certain parts of far this will just add even more confusion

viklequick:
> I think its better to make it universally but not only editor.. This feature of separate clipboard is not very easy to use itself, but if it will be also only in certain parts of far this will just add even more confusion

Got it, thank you. 

So, I've made:

* Options to operate with primary clipboard global, moving it from editor settings to interface settings;
* Added primary/clipboard buffers operational both in dialog controls;
* Added primary/clipboard buffers operational in command line.

Now in all cases as we have mouse selection (except special cases like screen box grabber) we have support of primary selection; and in pair cases we have selection from primary buffer as well.

Looks like this function is now consistent.

viklequick:
Looks like using OSC52 we can add support of both clipboards for TTY mode. Investigating... unfortunately, it needs addition to tty code.

viklequick:
The intermediate results of the experiments:

1. Using OSC52 Write (e.g. the approach we have in far2l for Ctrl+C if configured) we can specify the buffer easily.
2. For backward operation, we need OSC52 Read and this part of the OSC52 is very limited to xterm, kitty, alacritty, iTerm2 and few others. VTE-based terminals (like gnome-terminal) have no support of it. Konsole bans OSC52 Read for security.

Kitty --  Ask User -- needs to configure  `clipboard_control read-clipboard` in config
iTerm2  -- Toggle  Global Prefs -> Applications can access clipboard
Konsole --  Hard Block  Not currently possible without a patch
Ghostty -- Ask User  Enabled by default (Safe)

The bracketed paste (the way we currently have on far2l) have no option to ask the clipboard; but it can be configured like this way:

image

Still digging. Looks like we can add OSC52 Read as an option for users. I'll let you know about further progress.

There is a small application that can check the OSC52Read capability in your terminal:

```
#include <iostream>
#include <string>
#include <vector>
#include <termios.h>
#include <unistd.h>
#include <poll.h>

// --- Minimal Base64 Decoder (Required to read OSC 52) ---
std::string base64_decode(const std::string& in) {
    static const std::string table = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
    std::string out;
    std::vector<int> T(256, -1);
    for (int i = 0; i < 64; i++) T[table[i]] = i;
    int val = 0, valb = -8;
    for (unsigned char c : in) {
        if (T[c] == -1) break;
        val = (val << 6) + T[c];
        valb += 6;
        if (valb >= 0) {
            out.push_back(char((val >> valb) & 0xFF));
            valb -= 8;
        }
    }
    return out;
}

int main() {
    // 1. Set Terminal to Raw Mode (so we can read the response immediately)
    termios oldt, newt;
    tcgetattr(STDIN_FILENO, &oldt);
    newt = oldt;
    newt.c_lflag &= ~(ICANON | ECHO); 
    tcsetattr(STDIN_FILENO, TCSANOW, &newt);

    // 2. Send the OSC 52 Query for Clipboard ('c')
    // Format: ESC ] 52 ; c ; ? BEL
    std::cout << "\033]52;c;?\a" << std::flush;

    // 3. Read response with timeout using poll()
    std::string response;
    struct poll32 { struct pollfd p; } p32; // Wrapper for clarity
    struct pollfd fds[1];
    fds[0].fd = STDIN_FILENO;
    fds[0].events = POLLIN;

    bool found_response = false;
    // Wait up to 1000ms for the terminal to "type" the response back
    if (poll(fds, 1, 1000) > 0) {
        char buf[4096];
        ssize_t n = read(STDIN_FILENO, buf, sizeof(buf));
        if (n > 0) response.assign(buf, n);
    }

    // 4. Restore Terminal Mode
    tcsetattr(STDIN_FILENO, TCSANOW, &oldt);

    // 5. Parse the response
    // Typical response: \033]52;c;<base64>\a
    size_t header_pos = response.find("\033]52;c;");
    if (header_pos != std::string::npos) {
        size_t data_start = header_pos + 7;
        size_t data_end = response.find_first_of("\a\033", data_start);
        
        if (data_end != std::string::npos) {
            std::string b64 = response.substr(data_start, data_end - data_start);
            std::string decoded = base64_decode(b64);
            std::cout << "\n[Success] Clipboard Content: " << decoded << std::endl;
        }
    } else {
        std::cout << "\n[Error] No OSC 52 response received. Terminal might not support reading." << std::endl;
    }

    return 0;
}
```

Use it in this way:

```
g++ -o osc52req osc52req.cpp
./osc52req
```

elfmz:
conflicts..

viklequick:
Just to use as a reference:

Начнем издалека -- так получилось (с), что в X11 аж три разных клипборд буфера.

Первый знают все -- он так и называется "clipboard". Тот самый, который по шорткатам Copy/Paste.

Второй многие тоже знают -- называется primary и в него автоматом попадает то что выделено мышкой. Чисто "мышиный" такой карман -- выделение автоматом складывает, клик по средней кнопке вставляет.

И есть еще secondary, но им мало кто пользуется.

Фар-отец с windows очевидно умеет только один карман -- в окошках одного хватит всем.

Соответственно, когда мы в far2l стали добавлять поддержку нескольких клипборд буферов, то

* В GUI проблем нет, работает для редактора, ком строки, диалогов -- в обе стороны. Именно так как принято в X11. Если композитор поддерживает соответствующий протокол -- то и в Wayland.
* С терминалами ад, содом и далее по списку. Некоторые терминалы умеют OSC52 -- но каждый со своим видением "security".

Так что для терминальной версии реализовано вот что

* "туда" из фара в буфер -- работает и клипборд и праймари буферы, и выделение мышой и шорткаты будут работать независимо. Для этого в настройках интерфейса надо включить поддержку протокола OSC52 [Write].
* "обратно" из буфера в фар -- запрещено в целях безопасности. Протокол OSC52 Read выключен по умолчанию, иногда просто заблокирован (пример - konsole), так что нужен совместимый терминал и в нем отдельно разрешить чтение clipboard. Тогда будет как в гуях.

Так что без дополнительных настроек терминала, в терминале far2l не может сам запросить чтение клипборда и тем более не может указать что он хочет. Видеокамера в туалете работает для вашей безопасности (С). Но он может принять из терминала посылку -- которая и называется bracketed paste.

Так что если кто хочет чтобы работали оба клипборда в терминале -- настраивайте терминал. Типовой способ выглядит так

* Ctrl+V / Shift+Ins -- вставка из Clipboard buffer
* Ctrl+Shift+Ins - вставка из Primary buffer.

Но и тут есть ложка дегтя -- терминалы на движке VTE (тот же gnome terminal) Shift+Ins мапят на Primary buffer, и не дают поменять.

Вот такой он, мир терминалов.

Я добавил поддержку и OSC52 Read, только очень немного терминалов его разрешают. Поэтому я читаю с таймаутом, не пришло так не пришло.

elfmz:
терминал в far2l под far2l умеет копипастить если что и туда и сюда. Секурити обеспечивается ключом специфичным для конкретного клиента + совсем уж для гарантии тем, что терминал-хост отслеживает нажатие ctrl+C/Ctrl+ins и тп и разрешает копирование лишь после такового нажатия. Надежная армированная костылями конструкция вобщем.

viklequick:
> терминал в far2l под far2l умеет копипастить если что и туда и сюда. Секурити обеспечивается ключом специфичным для конкретного клиента + совсем уж для гарантии тем, что терминал-хост отслеживает нажатие ctrl+C/Ctrl+ins и тп и разрешает копирование лишь после такового нажатия. Надежная армированная костылями конструкция вобщем.

Это верно, я пока разбирался -- восхитился изящности решения. Крайне непростая вещь и при этом сделана удобно. 

Добавить OSC52 Read оказалось не так сложно как я думал.  Парсер научить ответу, добавить запрос на получение этого ответа, и аккуратно синхронизировать потоки записи и чтения по аналогии. И дает выбрать нужный буфер, я для того и полез :-)

Да, хотел посоветоваться. У нас в окне вывода в консоли есть свое выделение мышью, квадратное. Я пока его не трогал. Есть ли смысл (с отдельной опцией разумеется) переделать на классику по строкам? Оно не настолько мешает, но все-таки...

viklequick:
В сущности, это готовое решение -- везде где можно добавлено и подключено.  Я сам пользуюсь -- уже привык ;-)

viklequick:
@elfmz  если что, вся машинерия работает, опции я перенес, все замечания учел. Изменение полезное, таким аксакалам как я жизнь станет чуточку легче.

Если еще надо что подправить -- отпишитесь пожалуйста.

viklequick:
Added primary / clipboard buffers support for SDL backend

viklequick:
It is still ready to merge, by the way

PR #3158, opened 2025-12-19
File resume while copy

viklequick:
Initial version of file resume while copy

image


viklequick:
Added screenshot and l;ast minute fixes, loks like it is working as expected

PR #3168, opened 2025-12-28
Printer support in far editor

viklequick:
Added printer support in far editor through wxWidgets printing (via HTML/RTF) for UI, and simple lp-based for console. Printer caps are exposed by corresponding backend.

image

Using editor colors, the code computes "close" colors of the text to the white paper using RGB-LAB color space conversion. The character colors are being used from the editor directly, so if colorer didn't finished the colorization printer support will not see it too.

### DIFF
--- a/WinPort/src/Backend/Backend.h
+++ b/WinPort/src/Backend/Backend.h
@@ -307,15 +307,15 @@ struct WinPortMainBackendArg
 
 class IPrinterSupport {
 public:
-	virtual void PrintText(const std::wstring& jobName, const std::wstring& text) = 0;
-	virtual void PrintReducedHTML(const std::wstring& jobName, const std::wstring& text) = 0;
-	virtual void PrintTextFile(const std::wstring& fileName) = 0;
-	virtual void PrintHtmlFile(const std::wstring& fileName) = 0;
+	virtual void PrintText(const wchar_t* jobName, const wchar_t* text) = 0;
+	virtual void PrintReducedHTML(const wchar_t* jobName, const wchar_t* text) = 0;
+	virtual void PrintTextFile(const wchar_t* fileName) = 0;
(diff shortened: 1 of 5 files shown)


viklequick:
Actually, it is looks likee this way:

image

(standard preview for wxWidgets)

viklequick:
The same for macos, selected fragment

image

viklequick:
UTF-8 works as expected:

image

viklequick:
Macos native support

image

viklequick:
Basically, the code works as follows (wxWidgets backend active):

1. Examine selected fragment (or whole file) against colors that are being used for painting.
2. Using RGB -> LAB -> RGB conversion, colors are adjusted to be visible on white background (paper).
3. The result is converted to specific HTML (compatible with wxHtmlPrinting component)
4. Using wx UI, the file is displayed as preview, with ability to print it on printer directly.
5. For macos, the native components are being used instead of wx, but the HTML file is the same.

For terminal backend, as there is no UI available, the file is placed "as is", as text file, and then send to printer using lp command.  If  the printer configured, the file (or selected fragment) is being sent to the default printer (if configured). I've checked this against real hardware printer.

The HTML generation was not moved out to the backend on purpose: the same HTML can be used to support clipboard (with additional flavor) directly in editor.

viklequick:
Looks like it is ready to merge

elfmz:
Please avoid using of STL classes in Backend.h, as its defines interface across different binaries which potentially can be built with own allocators. So all that std::wstring in IPrinterSupport can be changed to const wchar_t * IMHO without any functional or aestetical problems

viklequick:
> Please avoid using of STL classes in Backend.h,

Got it, thank you. I've replaced std::wstring to wchar_t*.

Issue #1681, opened 2023-05-15
Mouse sensitivity parameter


valdemar1977:
Please add mouse sensitivity parameter for change mouse scrolling sensitivity, now it it very high in macOS. It is not possible to use mouse in file panels, text editor or menus.

PR #3210, opened 2026-01-25
Fix scrolling from precise input devices

exkrexpexfex:
such as macOS mouse or trackpad
Touch #1681

### DIFF
--- a/WinPort/src/Backend/WX/wxMain.cpp
+++ b/WinPort/src/Backend/WX/wxMain.cpp
@@ -1844,6 +1844,27 @@ void WinPortPanel::OnMouseNormal( wxMouseEvent &event, COORD pos_char)
 {
 	INPUT_RECORD ir = {0};
 	ir.EventType = MOUSE_EVENT;
+	auto enqueue_mouse_event = [&](INPUT_RECORD &rec) {
+		if (!(rec.Event.MouseEvent.dwEventFlags & MOUSE_MOVED)) {
+			fprintf(stderr, "Mouse: dwEventFlags=0x%x dwButtonState=0x%x dwControlKeyState=0x%x\n",
+				rec.Event.MouseEvent.dwEventFlags, rec.Event.MouseEvent.dwButtonState,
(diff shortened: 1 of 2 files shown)


elfmz:
can you elaborate on what was done?

exkrexpexfex:
We have received a number of comments on macOS scrolling sensitivity (though I never felt it inconvenient), and recently people reported same issue on wxGTK.
Here is what I found and what was changed:  
[WinPort/src/Backend/WX/wxMain.cpp:1860-1865](https://github.com/exkrexpexfex/far2l/blob/d10722638b62fbd0f8c18e496944960aa27e4614/WinPort/src/Backend/WX/wxMain.cpp#L1860-L1866) maps any non‑zero GetWheelRotation() to a single wheel event. This ignores both GetWheelDelta() and the magnitude of the rotation. On macOS, wx sends many small deltas per gesture, so each micro‑delta becomes a full scroll step, converted to arrow key press events by far2l core code, therefore the scrolling feels too sensitive.
[WinPort/src/Backend/WX/wxMain.cpp:1899-1905](https://github.com/exkrexpexfex/far2l/blob/d10722638b62fbd0f8c18e496944960aa27e4614/WinPort/src/Backend/WX/wxMain.cpp#L1899-L1905)explicitly bypasses the duplicate‑event filter for wheel events. That makes the high‑frequency macOS wheel stream unthrottled.
[wxWidgets-3.3.1/src/osx/cocoa/window.mm:807-849](https://github.com/wxWidgets/wxWidgets/blob/49c6810948f40c457e3d0848b9111627b5b61de5/src/osx/cocoa/window.mm#L807-L849) sets m_wheelDelta = 10 and uses scrollingDelta* directly for precise devices. This assumes consumers accumulate/scale deltas. far2l doesn’t, so it amplifies small deltas into big scroll jumps.

I have accumulated reported Wheel deltas to N character-cell steps using FontWidth() for horizontal and FontHeight() for vertical scrolling, which seems to be the right way for our character-grid terminal and shall not impact classic mouse wheels generating 120 per notch.

PR #3223, opened 2026-01-30
Show Pseudo-graphics on VT100+ terminals

SergeiPavlov:
NetBSD and other UNIX systems in console mode use DEC VT100-compatible terminal by default.
`TERM` environment variable is `wsvt25` (but it may be also `vt100`,  `vt220` ... on machines with real terminals, e.g. VAX).

The current `far2l` version prints any pseudo-graphics char as `â` (this is 0xE2 - the first byte of UTF-8-encoded box characters).

To show pseudo-graphic characters we have to switch to special codepage by `ESC (0` sequence
and then back by `ESC (B`.
The set of `VT100` graphics chars is limited: `┘┐┌└┼┤─├│┬┴`  ( https://vt100.net/docs/vt220-rm/chapter2.html#T2-4 ).
But we can approximate Far's double lines `═ ║` with them.

There is no problem in NetBSD GUI, where `TERM`=`xterm*` and it understands UTF-8.

### DIFF
--- a/WinPort/src/Backend/TTY/TTYOutput.cpp
+++ b/WinPort/src/Backend/TTY/TTYOutput.cpp
@@ -1,4 +1,3 @@
-
 #include <stdarg.h>
 #include <assert.h>
 #include <base64.h>


elfmz:
i added some changes for performance etc, can you test if its still (in current master( works fine for you?

SergeiPavlov:
> i added some changes for performance etc, can you test if its still (in current master( works fine for you?

I confirm the `master` branch shows graphics chars on VM-hosted  NetBSD but it loses many chars. 
And previous versions lose as well
Probably this is not related to VT100-specific logic, but is on some lower level.

I will try a real PC later.

SergeiPavlov:
@elfmz 
on the real PC:
The current state of the `master` does not work correctly on NetBSD (half of screen is lost)

[6063fe71ae34cef310e89320f20f7e1a386e23c9](https://github.com/elfmz/far2l/commit/6063fe71ae34cef310e89320f20f7e1a386e23c9) state does work

but commit https://github.com/elfmz/far2l/commit/51f568fbf309846de563238a70e6449a2a4eaf6a  breaks something

The fix: https://github.com/elfmz/far2l/pull/3229

PR #3241, opened 2026-02-05
Improve performance for large file editing and search operations

YuryChaikou-ServiceNow:
- Increase I/O buffer sizes from 8KB to 64KB for better read throughput
- Add bulk loading mode to skip expensive cache operations during file load
- Add regex caching to avoid recompilation of repeated search patterns
- Optimize case-insensitive search with pre-converted strings and first-character quick check

### DIFF
--- a/far2l/src/viewer.cpp
+++ b/far2l/src/viewer.cpp
@@ -2955,6 +2955,7 @@ void Viewer::Search(int Next, int FirstChar)
 							break;
 						}
 					}
+				}
 
 				if (ReverseSearch) {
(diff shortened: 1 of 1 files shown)


elfmz:
There were some embedded systems built with default stack size of 256K, however OldGetFileString is used as stack object and this such big can cause stack overflow in such systems:
```
char ReadBuf[65536];
wchar_t wReadBuf[65536];
```

So better make ReadBuf and wReadBuf to be std::vector

YuryChaikou-ServiceNow:
Done

PR #3256, opened 2026-02-11
Overallocation to fix AddOKCancel crash

exkrexpexfex:
due to realloc between OK and Cancel creation invalidating OK pointer.

Reported when using portable far2l built on musl via ssh when opening Options->interface settings, which contains in this case 33 elements so that reallocation happens between OK and Cancel buttons creation. Version built on glibc forgives this bug, musl - doesn't

See also modernization attempt with std::vector<std::unique_ptr<T>>  which fixes the issue by overcomplication instead of overallocation https://github.com/elfmz/far2l/compare/master...exkrexpexfex:far2l:reallocdialog

Shall we go with it instead? Please consider


elfmz:
please check if https://github.com/elfmz/far2l/pull/3274 fixes issue

exkrexpexfex:
@elfmz latest #3274 fixes the crash, thanks.

PR #3283, opened 2026-02-24
Fix far:config crash - more accurately process string without default

akruphi:
Additional to #3282


tempergate:
works correctly with #3282 and #3283

image

PR #3285, opened 2026-02-26
Added memo plugin - multi-page scratchpad for Far2l

stpork:
A multi-page scratchpad/memo plugin providing 10 independent memo pages with auto-save and keyboard navigation.

Features:
- Press Ctrl + S (Cmd + S) to launch
- 10 separate memo pages (Memo 1-10), each can be exported to file via F2/Shit+F2
- Full-featured multiline text editor (DI_MEMOEDIT)
- Keyboard navigation: Ctrl+0..9 / Alt+0..9 to switch memos
- Auto-save on switch, close, or memo change
- State persistence - restores last used memo

### DIFF
--- a/colorer/src/pcolorer2/FarEditorSet.cpp
+++ b/colorer/src/pcolorer2/FarEditorSet.cpp
@@ -730,10 +730,7 @@ int FarEditorSet::editorEvent(int Event, void* Param)
       case EE_GOTFOCUS: {
         if (!getCurrentEditor()) {
           editor = addCurrentEditor();
-          if (editor) {
-            return editor->editorEvent(EE_REDRAW, EEREDRAW_CHANGE);
-          }
-          return 0;
+          return editor->editorEvent(EE_REDRAW, EEREDRAW_CHANGE);
         }
         return 0;
       }
@@ -904,21 +901,9 @@ FarEditor* FarEditorSet::addCurrentEditor()
   EditorInfo ei {};
   Info.EditorControl(ECTL_GETINFO, &ei);
 
-  // Check if EditorID is valid (non-zero and not already tracked)
-  // This prevents adding editors for dialog controls like DI_MEMOEDIT
-  // which share context with the original editor behind the dialog
-  if (ei.EditorID == 0 || farEditorInstances.find(ei.EditorID) != farEditorInstances.end()) {
-    return nullptr;
-  }
-
-  // Check if we have a valid filename
-  const auto s = getCurrentFileName();
-  if (!s || s->isEmpty()) {
-    return nullptr;
-  }
-
   auto* editor = new FarEditor(&Info, parserFactory.get());
   farEditorInstances.emplace(ei.EditorID, editor);
+  const auto s = getCurrentFileName();
   editor->chooseFileType(s.get());
   editor->setTrueMod(useExtendedColors);
   editor->setRegionMapper(regionMapper.get());


unxed:
I definitely need it. Now I'm using new files like 1.txt for that

stpork:
Far-Memo

tempergate:
looks great

tempergate:
> * Press Ctrl + S (Cmd + S) to launch

control + S not working
memo pages , works with f11-> Memo
multiline works
Ctrl+0..9 / Alt+0..9 works
Auto-save works
State persistence works

tempergate:
have a debug is great, but, make an option to disable it for regular users, and make it turnable using for example state.ini parameter 

image

tempergate:
also, it could only called from file panels, probably it should be useful to make it available in all panels, in editor, viewer, and non-modals

stpork:
```
Memo Plugin – Summary of Changes

Crash Root Cause
DI_MEMOEDIT always creates an internal Editor. When focused, far2l
unconditionally broadcasts EE_GOTFOCUS (NotifyDialogEditorFocus,
dlgedit.cpp:214), regardless of DIF_EDITOR.

The colorer plugin handles EE_GOTFOCUS:
  editorEvent() → addCurrentEditor() → getCurrentFileName()

Because the dialog editor has no backing file, ECTL_GETFILENAME
returns nullptr. That nullptr is passed into UnicodeString inside
FarEditor::chooseFileType(), causing SIGABRT (KERN_INVALID_ADDRESS).

Note: Although addCurrentEditor() checks EditorID == 0, the crash
occurs earlier inside getCurrentFileName().

Temporary Workaround
In OpenPluginW():
  - Call EditorControl(ECTL_GETINFO)
  - If EditorID != 0 (real editor active), return INVALID_HANDLE_VALUE
  - Do not open the memo dialog

Prevents crash until colorer adds null-filename protection.

Configuration Support
- New exported ConfigureW()
  F11 → Plugins → Configure → Memo
- Dialog layout based on incsrch (tight margins, no inner separator,
  DIF_CENTERGROUP buttons, spacing around checkbox)
- Enabled flag stored in state.ini:
    Enabled=0|1
- Default: enabled (if key absent)
- OpenPluginW() checks flag first
  If disabled → Ctrl+S silently ignored

Localization
- Added:
    memo/lang/memoe.lng (English)
    memo/lang/memor.lng (Russian)
- MsgId enum:
    MMemo, MOk, MCancel, MConfigTitle, MEnablePlugin
- All labels via:
    g_far.GetMsg(g_far.ModuleNumber, MXxx)
- CMake copies .lng files to Plugins/memo/plug/
  (same pattern as incsrch)

Debug Logging
- DebugLog() → printf-style variadic + HH:MM:SS timestamp
- DBG(fmt, ...) macro (compiled out in Release via NDEBUG)
- Debug build log path:
    ~/.config/far2l/plugins/memo/debug.log (append)

Log coverage:
  1) Startup: module path, data dir, platform, build date, SysID, hotkey
  2) Key registration (macro-based)
  3) OpenPluginW invocation + rejection reason
  4) Ctrl/Alt+0–9 switching
  5) F2 / Shift+F2 export (paths + result)
  6) ESC close
  7) File I/O (path, byte count, failures)
  8) Dialog lifecycle (index, size, init failure)

UI Improvements
- Dialog title initialized immediately:
    [ Memo - N ]
  (no longer shows “[ Memo ]” first)
- Removed DIF_EDITOR from DI_MEMOEDIT
  (no effect; overlaps with DIF_SEPARATOR2 0x00020000)

Build Cleanup
- Disabled adb, mtp, img plugins (not available on macOS/ARM)
- Replaced if/endif with comments to avoid CMake errors

Documentation
- Added feature overview, keyboard table, known limitations
- Detailed explanation of colorer null-filename crash + workaround
- Storage layout and language file tables
- Build, hotkey setup, and debug logging sections
```

tempergate:
<s>Critical
Brakes Colorer plugin for independent editor instances 
https://github.com/elfmz/far2l/issues/3290</s>

tempergate:
still not working

image

[far2l.log](https://github.com/user-attachments/files/25728682/far2l.log)

could we just return back highlighting in memo? It wasn't so bad idea to have it in memo

stpork:
I probably found the way to fix it gracefully. At least it works for me on Mac now in both editors with colorer.

tempergate:
> I probably found the way to fix it gracefully. At least it works for me on Mac now in both editors with colorer.

still not working correctly

[far2l.log](https://github.com/user-attachments/files/25730746/far2l.log)

stpork:
Fixed Memo plugin to avoid being invoked in editors when Colorer is enabled.

The Cause:
- DI_MEMOEDIT Nature: In far2l, this control is a full instance of the internal Editor class.
- Event Collision: Upon initialization, the control broadcasts standard editor events (e.g., EE_READ).
- Colorer Failure: As a global listener, Colorer intercepts these events. It attempts to determine the "file type" for highlighting, but because a dialog control lacks a physical file path, its FileName metadata is null. Colorer’s chooseFileType function dereferences this null pointer without validation, causing the crash.

If colorer would fix this defect I can enable Memo to work in editors too.

tempergate:
fix works a28b8f2a5152217742c01e94a002e910a1eb3ccd

akruphi:
* **Serious BUG**: Crash when inside Editor open any dialog (for example Search via **F7**) and press **Ctrl+S**

* Please make disabled by default `~/.config/far2l/plugins/memo/debug.log` (inside my system `-DCMAKE_BUILD_TYPE=Release` is not equivalent to `-DNDEBUG`)

* Incorrect configuration dialog width in Russian: 

@stpork

PR #3292, opened 2026-03-08
Fix: prevent editor crash on Ctrl+S and refactor settings logic

stpork:
Ctrl+S Crash Fix: 
Implemented a safety check that scans the Far2l window stack. The Memo dialog is now blocked if an Editor instance is already active, preventing a re-entrancy conflict in the Colorer plugin that previously led to a system crash.

Debug Logging (DBG) Logic:
Inverted the debug logic to use DEBUG instead of NDEBUG for compilation. Disabled logging by default.

Russian Dialog & Localization: Refined 

----
Issue with Colorer: Recursive Editor Instantiation Crash in Far2l

The Problem: 
A segmentation fault occurs when Ctrl+S (Memo plugin) is pressed inside the Far2l Editor. The crash is caused by a re-entrancy conflict in the Colorer plugin.

Root Cause: 
Nested Contexts: Ctrl+S opens a modal dialog with a DI_MEMOEDIT control. In Far2l, this is a full editor instance, not a simple text field.

Event Collision: Initializing DI_MEMOEDIT triggers an EE_READ event. Colorer intercepts this to provide highlighting for the new "mini-editor."

State Conflict: Colorer uses global pointers and shared caches for parse trees. Attempting to handle a new editor widget while a primary editor session is active in the background causes a global state conflict.

Memory Access: Colorer tries to apply background highlighting rules to the foreground dialog, leading to out-of-bounds memory access and an immediate crash.


akruphi:
Thank you for quick fixes.

> Debug Logging (DBG) Logic:
Inverted the debug logic to use DEBUG instead of NDEBUG for compilation. Disabled logging by default.

Inside PR I do not see any changes Debug logic - now `NDEBUG` continue processing in `memo.cpp` and mention in `README.md`. May be you forgot unpushed commits?

Additional ideas:
* Inside any memo processing **F9** to quick open Memo Plugin Configuration dialog.
* If possible not save empty `memo-00.txt`, ... `memo-09.txt` files. Ideal situation: save `memo-*.txt` **only if inside memo was changes** (now it always saved).

stpork:
@akruphi please check #3293

PR #3293, opened 2026-03-09
Rewritten memo plugin to work with F4 Editor without crash and use mouse-clickable FAR buttons to switch memo pages

stpork:
1. Memo can be used in F4 Editor without crash
2. Editors are not crashing anymore with Memo fix to "far2l/src/plug/plugapi.cpp"
3. Memo pages are working with buttons and can be clicked by mouse in addition with Ctrl/Alt+0..9
4. F9 Config can be invoked directly from Memo.
5. Autosave overwrites only if there are real changes
6. DBG uses explicitly defined _DEBUG or DEBUG instead of implicit NDEBUG earlier
7. Memo dialog auto-resizes with far2l window


memo/src/memo.cpp
```diff
@@ -1,67 +1,40 @@
...
+// Returns base directory for memo files, creates if missing
+static std::wstring GetMemoDir() {
+  const char *home = getenv("HOME");
+  std::wstring dir = home ? MB2Wide(home) + L"/.config/far2l/plugins/memo" : L"memos";
+  if (home) {
+    WINPORT(CreateDirectory)((MB2Wide(home) + L"/.config").c_str(), NULL);
+    WINPORT(CreateDirectory)((MB2Wide(home) + L"/.config/far2l").c_str(), NULL);
+    WINPORT(CreateDirectory)((MB2Wide(home) + L"/.config/far2l/plugins").c_str(), NULL);
+    WINPORT(CreateDirectory)(dir.c_str(), NULL);
+  }
```
akruphi: Far2l do not always use `~/.config/far2l` to save settings. Location may be override via command line switch `-u` or via `FARSETTINGS` environment variable (see info in build in help via **F1** in section **FAR2L features - Getting Started**). Please don't use `HOME` environment for settings location.

Inside far2l code you should use `InMyConfig("",FALSE).c_str()` to obtain config directory (`FALSE` here to get directory name without generate it).

In your plugin you should use
```cpp
InMyConfig("plugins/memo/"); // get directory name and generate directory (if it not exist)
InMyConfig("plugins/memo/", FALSE); // only get directory name but don't check it existing.
```

stpork: Thanks for the hint. Updated in 8ad31e7c3d2f524263bc980e75c7b1ef6f0cb228 among other small enhancements

akruphi:
**BUG:** I compile **8ad31e7c** and any change memo via **Ctrl+1** ... **Ctrl+9** make far2l **crash** :crying_cat_face:.

memo/lang/memoe.lng
```diff
@@ -3,6 +3,11 @@
 "Memo"
 "&OK"
 "&Cancel"
-"Memo Plugin Configuration"
-"&Enable Memo Plugin"
-"&Use Ctrl+S to open"
+"Memo Plugin"
+"Enable &Memo plugin"
+"Enable Ctrl+&S hotkey"
```
akruphi: May be explicit write here that hotkey work via common macros:
`"Enable Ctrl+&S hotkey (via common macros)"`

memo/lang/memor.lng
```diff
@@ -1,8 +1,13 @@
...
+.Language=Russian,Russian
 
 "Memo"
 "&OK"
 "&Отмена"
-"Настройка плагина Memo"
+"Настройка Memo"
 "Включить плагин &Memo"
-"Вызывать через &Ctrl+S"
+"Использовать Ctrl+&S"
```
akruphi: `"Использовать Ctrl+&S (через общий макрос)"`

memo/src/memo.cpp
```diff
@@ -1,654 +1,402 @@
...
+      kfh.Save();
+
+      ActlKeyMacro akm = {MCMD_LOADALL};
+      g_far.AdvControl(g_far.ModuleNumber, ACTL_KEYMACRO, &akm, NULL);
+    }
   } else {
-    return; // No change
-  }
+    if (kfh.HasSection(MACRO_SECTION)) {
+      kfh.RemoveSection(MACRO_SECTION);
```
akruphi, line 201: This unthinking deletion is not correct behavior if user has own **CtrlS** macros for other purpose.

You should check `MACRO_K_SEQ` and do not delete if `MACRO_K_SEQ` is not equivalent your macros and please also  show warning message and make option disabled.

memo/src/memo.cpp
```diff
@@ -1,654 +1,402 @@
...
-  DBG("startup: hotkey   = Ctrl+S / Cmd+S (via key_macros.ini CallPlugin)");
-  DBG("startup: prefix   = \"memo\" (command line prefix)");
-  DBG("startup: SYSID    = 0x%08X", SYSID_MEMO);
-
-  // Auto-register macro if enabled
-  if (IsHotkeyEnabled()) {
+  g_far = *Info; InMyConfig(MEMO_DIR);
+  KeyFileHelper kf(InMyConfig(INI_PATH, false));
+  g_enabled = kf.GetInt(INI_S_SETTINGS, INI_K_ENABLED, 1);
+  g_useHotkey = kf.GetInt(INI_S_SETTINGS, INI_K_HOTKEY, 1);
```
akruphi, line 496: Because hotkey work via macros and it may be conflict if user has own **CtrlS** macros I suggest make it disabled by default:
```
g_useHotkey = kf.GetInt(INI_S_SETTINGS, INI_K_HOTKEY, 0);
```

stpork: I added a check for an existing Ctrl+S macro that may differ from the one calling Memo.

1. If another Ctrl+S macro is detected at startup, the plugin does not overwrite it automatically and instead sets a flag.
2. If the existing Ctrl+S macro is different from ours, the plugin does not remove it when the configuration flag is disabled.
3. However, if the existing Ctrl+S handler is different and the user explicitly enables Ctrl+S in the plugin settings, the plugin overwrites the section. The previous key_macro.ini is backed up as *.bak (*-1.bak, *-2.bak, etc., if files with the same name already exist).

See  0b75a8ccd2bf388ff9a2b09cee9950f54c2068eb

akruphi:
Thank you for correction macros logic. I continue advice explicit write in config checkbox-text about work via **common macros** and make this feature off by default.

**BUG:** In commit **0169a6c9** far2l continue crash when try change from current to any other memo.

stpork:
I fixed bug, and added (via common macros). What comes to disabled by default - the plugin checks if there is another macros handler installed for Ctrl+S before changing the hotkey to itself. If there is already some Ctrl+S - it will not overwrite. 

commit 57e21f2de5c62395680033417d17ee0f1909a2e9

akruphi:
In 57e21f2d no crash anymore.

Thank you for fix bug and add explanation text near checkbox.

Minor inconsistent: when I switch to other memo plugin generate empty `memo-0?.txt` file even when I did not write any changes in memo.

akruphi:
Because **Esc** always save changes I suggest inside memo add new button with name `[ Reset to saved ]` or `[Discharge changes]` with key **Ctrl+R**.

memo/lang/memoe.lng
```diff
@@ -3,6 +3,12 @@
...
+"F2 Save"
```
akruphi, line 9: May be write as `"F2 Export to file"`?

memo/lang/memor.lng
```diff
@@ -1,8 +1,14 @@
...
+"F2 Сохранить"
```
akruphi, line 9: `"F2 Экспорт в файл"`

stpork:
I added normalization to prevent creating empty files.

Regarding save/discard/reload: Memo is meant to be a small scratchpad with 10 pages that always keep the latest edit state, similar to many note apps. The goal is quick notes without creating random files like 1.txt in far2l. It was never intended to be a full editor with rich save/export functionality, and I would prefer to avoid implementing more complex state logic.

About the F2 strings: I would like to keep them as short as possible. There is no actual “save” in Memo (all 10 pages are persistent anyway), so calling it “Export” feels somewhat misleading since nothing is really being saved in plugin anyway.

atolismesh:
Why did you choose Ctrl-S as your shortcut? It's reserved for command line navigation when panels are enabled. I use it every day. Now I have to constantly disable this plugin.

stpork:
@atolismesh I've changed default hotkey to Ctrl+Shift+S to address your problem. Hopefully that one is not occupied. The logic behind was - the combination should be handy enough to be pressed with one hand without changing the "standard" keyboard palm position.

atolismesh:
@stpork Thanks! Sounds great!

akruphi:
> I've changed default hotkey to Ctrl+Shift+S 

In Linux Desktop Environment if **Ctrl+Shift** used for switch language **Ctrl+Shift+any_keys** do not obtaining in applications.

I recommend you make any additional macros initial **disabled by default**, but in configuration add several variants for user's choice for **Ctrl+S**, **Ctrl+Shift+S**, **Alt+Shift+S**, etc.

For example in OpenWith plugin we did not add keys silently, but in help in section **Tips & tricks** describe detailed advice to user about key-macros: https://github.com/elfmz/far2l/blob/4cd53660711bab309d8c3de0446cfc8362c7f791/OpenWith/configs/help_en.hlf#L193-L215 and https://github.com/elfmz/far2l/blob/4cd53660711bab309d8c3de0446cfc8362c7f791/OpenWith/configs/help_ru.hlf#L193-L215

UPD: in any cases your plugin easy obtain from Panels/Command line/Viewer/Editor/Dialogs via **F11**->Memo without any additional macros.

stpork:
I've changed to Ctrl + Alt + S in order to not build the complexity around the simple logic (at least for the time being).

akruphi:
> I've changed to Ctrl + Alt + S in order to not build the complexity around the simple logic (at least for the time being).

Please also change in **README.md** all `Ctrl+S` to `Ctrl+Alt+S`

PR #3295, opened 2026-03-10
Fix Delete to Trash on macOS

YuryChaikou-ServiceNow:
The trash helper treated every "trash" binary like trash-cli and invoked "trash put <path>". On macOS, "/usr/bin/trash" expects files directly, so far2l tried to trash a literal file named "put" and every Delete-to-Trash operation failed with a bogus file-not-found error. Detect the macOS CLI signature and use the correct invocation while preserving the existing trash-cli path.


elfmz:
is this 'grep fileToMoveToTrash' really reliable? May be better just to check that is macos using something like uname?

YuryChaikou-ServiceNow:
Good point. Using uname now.

PR #3298, opened 2026-03-11
Improve performance and UI responsiveness

YuryChaikou-ServiceNow:
Input latency:
- Reduce idle WaitConsoleInput timeout from 160ms to 50ms
- Increase KEY_IDLE delivery frequency from 1000ms to 250ms

Rendering:
- Cache color state in SetColor/SetFarColor to skip redundant Console.SetTextAttributes() calls when color is unchanged
- Build symlink display string forward (Append) instead of double Insert(0,...) in ConvertName

O(n²) → O(n) string operations:
- FormatNumber: build digit string left-to-right with separator logic instead of Insert(0, char) in reverse loop
- FixedSizeStr: build right-aligned string forward (spaces then content) instead of Insert(0, spaces) shifting existing content
- StrCellsTruncateLeft: incremental cell counting instead of calling StrCellsCount() per offset
- StrCellsTruncateCenter: cache left/right cell counts and update incrementally instead of recomputing O(n) per iteration

TTY backend:
- Replace tcdrain() on every writer cycle with conditional call only before going to background
- Replace 12 notify_all() calls on _async_cond with notify_one() since WriterThread is the sole waiter; keep notify_all() only for shutdown paths (deadio, exiting)

NetRocks threading:
- Replace ProgressStateUpdate 1-second usleep polling loop with condition_variable wait_for(500ms), notified on pause/abort state changes
- Add std::condition_variable to ProgressState and notify at all 5 sites that modify paused/aborting flags

ThreadedWorkQueue:
- Release mutex before sleep(1) during OOM recovery to avoid blocking all worker threads
- Use notify_one() instead of notify_all() for completion

### DIFF
--- a/NetRocks/src/Op/Utils/ProgressStateUpdate.cpp
+++ b/NetRocks/src/Op/Utils/ProgressStateUpdate.cpp
@@ -9,19 +9,15 @@ ProgressStateUpdate::ProgressStateUpdate(ProgressState &state)
 {
 	std::chrono::milliseconds pause_ticks = {};
 
-	for (;;) {
-		if (state.aborting)
-			throw AbortError();
-		if (!state.paused)
-			break;
-
-		unlock();
+	while (state.paused && !state.aborting) {
 		if (pause_ticks.count() == 0) {
 			pause_ticks = TimeMSNow();
 		}
-		usleep(1000000);
-		lock();
+		state.cond.wait_for(*this, std::chrono::milliseconds(500));
 	}
+
+	if (state.aborting)
+		throw AbortError();
 
 	if (pause_ticks.count() != 0) {
 		pause_ticks = TimeMSNow() - pause_ticks;
--- a/NetRocks/src/Op/OpBase.cpp
+++ b/NetRocks/src/Op/OpBase.cpp
@@ -115,6 +115,7 @@ void OpBase::ForcefullyAbort()
 	{
 		std::lock_guard<std::mutex> locker(_state.mtx);
 		_state.aborting = true;
+		_state.cond.notify_all();
 	}
 
 	_base_host->Abort();
--- a/NetRocks/src/Op/OpXfer.cpp
+++ b/NetRocks/src/Op/OpXfer.cpp
@@ -157,6 +157,7 @@ void OpXfer::Abort()
 		if (_state.finished)
 			return;
 		_state.aborting = true;
+		_state.cond.notify_all();
 	}
 
 	_dst_host->Abort();
--- a/NetRocks/src/UI/Activities/AbortOperationRequest.cpp
+++ b/NetRocks/src/UI/Activities/AbortOperationRequest.cpp
@@ -133,12 +133,14 @@ void AbortOperationRequest(ProgressState &state, bool force_immediate)
 	if (!AbortConfirm().Ask()) { //param1 == _i_cancel &&
 		std::lock_guard<std::mutex> locker(state.mtx);
 		state.paused = saved_paused;
+		state.cond.notify_all();
 		return;
 	}
 	{
 		std::lock_guard<std::mutex> locker(state.mtx);
 		state.aborting = true;
 		state.paused = saved_paused;
+		state.cond.notify_all();
 	}
 	if (force_immediate && state.ao_host) {
 		state.ao_host->ForcefullyAbort();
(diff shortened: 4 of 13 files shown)


NetRocks/src/Op/Utils/ProgressStateUpdate.cpp
```diff
@@ -9,20 +9,16 @@ ProgressStateUpdate::ProgressStateUpdate(ProgressState &state)
...
+	while (state.paused && !state.aborting) {
 		if (pause_ticks.count() == 0) {
 			pause_ticks = TimeMSNow();
 		}
-		usleep(1000000);
-		lock();
+		state.cond.wait_for(*this, std::chrono::milliseconds(500));
 	}
 
+	if (state.aborting)
```
elfmz, line 19: why?

YuryChaikou-ServiceNow: The original code did `unlock(); usleep(1000000); lock();` - a manual unlock/sleep/relock loop with 1-second granularity. This means unpausing or aborting could take up to 1 second to be noticed.

Replaced with `condition_variable::wait_for()` which:
1. Wakes immediately when another thread calls `cond.notify_all()` (already done at all sites that modify paused/aborting)
2. Has a 500ms timeout as a safety net for spurious wakeups
3. Handles mutex release/reacquire atomically (cleaner than manual unlock/lock)

Net effect: pause/abort responsiveness improves from up to 1s to near-instant.

yulian5:
Review from Gemini 3.1 Pro:

Based on the details from the repository, **PR #3298 ("Improve performance and UI responsiveness")** in `elfmz/far2l` is a highly impactful set of optimizations by YuryChaikou-ServiceNow. 

Here is a review and breakdown of the changes, their implications, and the ongoing discussion:

### 1. Algorithmic Optimizations (Excellent)
* **O(n²) to O(n) String Operations:** This is the most objectively necessary fix in the PR. The original code was apparently using `Insert(0, ...)` inside loops for functions like `FormatNumber` and `FixedSizeStr`. Because inserting at the beginning of a string requires shifting all subsequent characters in memory, doing this in a loop results in $O(n^2)$ complexity. Rebuilding strings incrementally left-to-right (or right-aligned with spaces first) fundamentally speeds up UI formatting.
* **Truncation Caching:** Caching the cell counts in `StrCellsTruncateCenter` instead of blindly recomputing them per-iteration saves heavy repeated string traversals.

### 2. Thread Synchronization & TTY (Solid Improvements)
* **`notify_one()` vs `notify_all()`:** Replacing `notify_all()` with `notify_one()` on `_async_cond` (since `WriterThread` is the sole waiter) and ThreadedWorkQueue completions is a textbook best practice. It prevents the "thundering herd" problem where multiple threads wake up, compete for a lock, and immediately go back to sleep.
* **Polling to Event-Driven (`condition_variable`):** Replacing 1-second `usleep` polling with `std::condition_variable` drastically reduces idle CPU usage while making the UI more responsive to state changes. 
* **OOM Recovery Mutex:** Releasing the mutex before `sleep(1)` during Out-Of-Memory (OOM) recovery is a critical bug fix. Previously, a single thread hitting an OOM state would freeze all other worker threads while sleeping.

### 3. Rendering and Input Latency (Good Quality-of-Life)
* **Color Caching:** Skipping redundant `Console.SetTextAttributes()` calls when the color hasn't actually changed prevents unnecessary system calls, which are often expensive in terminal emulators.
* **Lowering Input Timeouts:** Decreasing `WaitConsoleInput` from 160ms to 50ms and increasing `KEY_IDLE` delivery will make the UI feel notably snappier to users, especially for fast typists or repetitive keystrokes.
* **`tcdrain()` Conditionally:** Moving `tcdrain()` (which blocks until all output has been transmitted) off the main writer cycle and only running it before going to the background stops the writer thread from unnecessarily stalling.

### Review of the Ongoing Discussion
The repository owner, `elfmz`, asked for clarification (**"why?"**) on a specific change in `NetRocks/src/Op/Utils/ProgressStateUpdate.cpp`:
```cpp
state.cond.wait_for(*this, std::chrono::milliseconds(500));
```

**The author's response is architecturally sound:**
The old code relied on a clunky `unlock(); usleep(1000000); lock();` loop. If a user tried to pause or abort an operation, it could take up to a full second for the thread to wake up and register the action. By switching to `wait_for`, the thread now sleeps but will instantly wake up if another thread calls `notify_all()` (e.g., when the user hits "abort"). The 500ms timeout simply serves as a fallback/safety net for spurious wakeups while the condition handles the lock release/reacquisition atomically.

### Overall Verdict
**Status: Approve & Merge**
This is a high-quality pull request. The author identified multiple classic performance bottlenecks (reverse string insertions, sleep-based polling loops, redundant system calls) and replaced them with standard C++ best practices. There are no obvious regressions, and the changes to the concurrency model should heavily reduce UI stuttering when handling heavy file operations.

PR #3300, opened 2026-03-11
Fix Find Files dialog freezing and improve responsiveness

YuryChaikou-ServiceNow:
Dialog responsiveness:
- Inject NOOP_EVENT periodically from worker threads and search thread to wake the modal dialog event loop, ensuring progress display updates and ESC key checks even when no user input arrives
- Add KickDialogIfNeeded() called from ScanFileByMapping (per mmap window), WorkProc completion, and DoScanTree loop
- Add ThreadedWorkQueue::SetWaitCallback() so the dialog gets kicked even while Queue() or Finalize() blocks waiting for workers

Progress display:
- Update displayed filename from worker threads (SetFindMessageMB) so the dialog shows which file is currently being scanned, not a stale name from the search thread blocked in Queue()
- Pass display name into ScanFileByMapping to update filename in sync with percent on each mmap window slide

### DIFF
--- a/far2l/src/findfile.cpp
+++ b/far2l/src/findfile.cpp
@@ -82,7 +82,7 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 // mmap'ed window size limit, must be multiple of any sane page size (0x1000 on intel)
 #if defined(__LP64__) || defined(_LP64)
-#define FILE_SCAN_MMAP_WINDOW 0x1000000
+#define FILE_SCAN_MMAP_WINDOW 0x100000
 #else
 #define FILE_SCAN_MMAP_WINDOW 0x10000
 #endif
@@ -2356,10 +2356,6 @@ static void DoPrepareFileList(HANDLE hDlg)
 		strRoot = pwRoot;
 		DoScanTree(hDlg, strRoot);
 	}
-
-	if (StopFlag && pWorkQueue) {
-		pWorkQueue->Abort();
-	}
 }
 
 static void DoPreparePluginList(HANDLE hDlg)
@@ -2395,10 +2391,6 @@ static void DoPreparePluginList(HANDLE hDlg)
 			|| SearchMode == FINDAREA_INPATH) {
 		PluginLocker Lock;
 		CtrlObject->Plugins.SetDirectory(ArcItem.hPlugin, strSaveDir, OPM_FIND);
-	}
-
-	if (StopFlag && pWorkQueue) {
-		pWorkQueue->Abort();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


elfmz:
How to reproduce issue that is fixed here?
Also can it be that it can be fixed by applying SetRegularIdle(true); on dialog?
Also i remember that there re lot of race condition in that code left until need to be fixed, may be they re true reason for that?

elfmz:
regarding increasing mmap window - i checked that values and chosen this as increasing more didnt give any measurable benefit (see fprintf(stderr, "FindFiles complete printout) However far2l sometimes used on embedded system with low amount of RAM, like routers, where using 16MB for mmap can be too much.

elfmz:
please check if still relevant after fd20472ab0b5afb281337f97f783050bb2ccf8c5

YuryChaikou-ServiceNow:
Tried on the latest master - could not reproduce the ESC-hang issue, reverted those changes.
Removed increase of mmap window as well.

To repro progress display issue:

1. Open Find Files, enter a content search pattern
2. Search a directory with many large files - the files need to be large enough that worker threads take noticeable time scanning each one
3. Observe the status line - it shows which file is being searched and a percentage
  
Without the fix:
  - The filename is only set by the search thread (itd.SetFindMessage() in DoScanTree), which shows the last file it queued, not the file workers are actually scanning
  - When the search thread blocks in Queue() (backlog full), the displayed filename goes stale - it stays on whatever file was last queued, even as workers move through other files
  - Without NOOP_EVENT injection, the dialog may not redraw at all if no user input arrives - so even the percentage can appear frozen

With the fix:
  - Workers call itd.SetFindMessageMB() to show the file they're currently scanning
  - KickDialogIfNeeded() injects NOOP_EVENT every ~200ms to force dialog redraws
  - SetWaitCallback ensures kicks continue even while Queue() is blocking
  
Regarding SetRegularIdle(true): It helps deliver idle events, but alone it fires only every 1000ms and doesn't cover the case where the search thread is blocked in Queue() waiting for backlog space - workers are scanning but the displayed filename is stale because only the search thread was updating it. The changes here add NOOP_EVENT injection at ~200ms intervals and let worker threads update the displayed filename directly.

elfmz:
'last file' is ok as there is no 'current file' - there re multiple workers scanning in parallel multiple files
Dlg.SetRegularIdle(true); since last change - also injects events with similar rate
in general its not very good idea to kick something from workers as they designed to be parallel independed threads, primarily for faster search, may be by cost of little bit worsenet reaction on UI - search speed is more important

elfmz:
Note that in general architecture of findfile threads is looking so:
Main FAR thread -> Find files thread (FindFileThread) -> multiple workers threads running in parallel (ScanFileWorkItem)


UI is running in main thread that is not blocked anyhow, so there is no need to kick it explicitly. It updated from time to time by timer but not per each event primarily for a sake of reduce losses on screen updates, and also due to actually having multithreaded file search means there is no obvious 'current file', just last queued one.

Issue #3171, opened 2025-12-30
vt::ExecuteCommandBegin fails because of motd being displayed


DamianFekete:
On some Linux distributions (like https://github.com/ublue-os/aurora) a motd (message of the day) is displayed by default when users log in.
Aurora uses [glow](https://github.com/charmbracelet/glow) to render a Markdown message.

`/etc/profile.d/ublue-motd.sh` executes this `ublue-motd` script:
```
#!/usr/bin/env sh

MOTD_TEMPLATE_FILE="${MOTD_TEMPLATE_FILE:-/usr/share/ublue-os/motd/template.md}"
. "${MOTD_ENV_SCRIPT:-/usr/share/ublue-os/motd/env.sh}"

envsubst < "${MOTD_TEMPLATE_FILE}" | glow -
```

It seems, that if `glow` is executed by `.bashrc`, the command that I try to execute in FAR is not executed at all in the BASH shell. I land in the shell and can type commands there, and I have to type `exit` to get back to FAR. Nothing is visible (neither the command or what I type).

For the following screenshot I've commented out https://github.com/elfmz/far2l/blob/b9702f965384b8b227e6e7c0da0fa62c833ec9a9/far2l/src/vt/vtshell.cpp#L1005 to be able to see what is going on. Please ignore the `{xxx}` blocks in the motd.

<img width="1031" height="817" alt="Image" src="https://github.com/user-attachments/assets/6e8212e6-b0a6-4bbd-a519-f7e128eb5d86" />

Everything works as expected after I disable the motd, but I was wondering if this issue could be better handled (either by FAR or glow).

---

The temp `vtcmd` files are not deleted as far doesn't find the echoed marker.

<img width="1133" height="436" alt="Image" src="https://github.com/user-attachments/assets/ea2d64c9-4811-4e90-9b84-1489b8777f3f" />

unxed:
Try #3337 please

PR #3337, opened 2026-04-05
[WIP] Fix shell initialization race condition with MOTD

unxed:
Improve VT shell startup by draining initial output before command injection. Previously, a fixed 300ms delay often proved insufficient if MOTD tools like glow were executed in .bashrc, leading to command buffer flushing or premature input loss. This change replaces the static sleep with an active wait loop that monitors terminal output and process group status, ensuring the shell is fully initialized and idle before sending commands.

Touch #3171


elfmz:
unrelevant now, see fix above - it much less dependent on timings as it establishes communication with shell, also in case having shell without such problems initialization now is instant

PR #3248, opened 2026-02-07
Git Gutter plugin

exkrexpexfex:
image

Experimenting with multiline colored edit over colored editor.

The plugin adds a vertical gutter column with colored marks showing how the current editor buffer differs from a chosen git baseline (HEAD by default; index/unstaged/branches selectable in  GitGutter options). Clicking a mark opens a popup diff (colored according to change type) with prev/next navigation and a Revert button. Revert applies changes via editor operations. Colors are configurable in the GitGutter config dialog. If the current file is not in a git repository, the plugin does nothing.

PR #3359, opened 2026-04-21
Editor: add vertical block typing and replacement

### LINKED ISSUE #1459: [feature request] vertical block column Editor, like Visual Studio Code
It would be great if Editor will be able to edit vertical blocks - typing text and/or pressing Backspace should be applied to each row of the vertical block.

Example is shown here (it's animated gif, please press Play button there to show the effect):

Also it would be nice to implement the multiple cursors as well, as explained there: https://code.visualstudio.com/docs/editor/codebasics

exkrexpexfex:
Closes #1459


akruphi:
Is new behavior not depended on `Flags.Check(FEDITLINE_OVERTYPE)`?

May be describe this feature in help in section `@Editor` in "Notes"?
For example (draft fantasy):
```
    4. If a vertical block is selected any text typing will replace the selected in all selected lines;
this behavior independent on #Ins# (Insert/Overtype) state.
```
```
    4. Если выбран вертикальный блок, любой набираемый текст заменит выделенное во всех выбранных строках;
это поведение не зависит от состояния #Ins# (Вставка/Замена).
```

exkrexpexfex:
The new behavior respects overtype and persistence blocks options. I've added a note in the help please check

tempergate:
good "quality of life" feature

image

<s>but  it works strange, when it is starting from an empty place or a column with no symbols before (265-269)
correct when starting from the same columns after existing symbols or replacing already existing (270-272)

it would be better to add spaces before it</s>

fixed

exkrexpexfex:
> works strange, when it is starting from an empty place or a column with no symbols before (265-269) correct when starting from the same columns after existing symbols or replacing already existing (270-272)

@tempergate thanks for testing, fixed, please check.
Also 0-width vertical blocks are supported, select with alt+up/down and start typing, please give it a try as well.

tempergate:
> > works strange, when it is starting from an empty place or a column with no symbols before (265-269) correct when starting from the same columns after existing symbols or replacing already existing (270-272)
> 
> @tempergate thanks for testing, fixed, please check. Also 0-width vertical blocks are supported, select with alt+up/down and start typing, please give it a try as well.

works great

far2l/src/editor.cpp
```diff
@@ -66,6 +66,67 @@ static int ReplaceMode, ReplaceAll;
...
+	if (end > length) {
+		end = length;
+	}
+	if (end <= start) {
+		line->SetCurPos(start);
+		return false;
+	}
+
+	const int eol_length = StrLength(end_seq);
+	wchar_t *tmp = new (std::nothrow) wchar_t[length - (end - start) + eol_length + 1];
```
elfmz: Why raw pointer and that wmemcpy with manual length calculations, but not std::wstring tmp; and tmp.append(cur_str, start) ?
And use tmp.reserve(length - (end - start) + eol_length + 1) to avoid reallocations. This will be more readable and in many cases faster than such manual sundown. And of course much less risk of occasional leg shooting.

tempergate:
@exkrexpexfex , there is a new conflict after editor patches, same as for #3248

far2l/src/editor.cpp
exkrexpexfex: Done

exkrexpexfex:
@tempergate thanks, updated

PR #3363, opened 2026-04-23
SDL: workaround for opengl driver and composer in KDE Plasma

viklequick:
The problem: once the accelerated backend started, it asks KDE Plasma to switch off the kwin_x11 compositor. Result the display flickering at the start.

Actually, as there is no video player and no game, we do not need to turn off the Plasma compositing at the start.

So, I've added hint for SDL to not play with kwin.

In addition, I've disabled unexpected SDL_MaximizeWindow as we already creating the window with the appropriate size.


tempergate:
Works under every composer, not only in KDE Plasma
rename "SDL: workaround for opengl driver and composer in KDE Plasma" to "SDL: workaround for opengl driver and composers"


image

PR #3373, opened 2026-04-28
ADB plugin: Corrections  for F5/F6 behavior — on-device ops, renames, overwrites, fixes

stpork:
- Shift+F6: rename in place; Shift+F5: duplicate (on-device mv/cp)
- F5/F6 same device: on-device copy/move, no host roundtrip;
  cross-mount paths fall back to on-device cp -a + rm -rf
- Overwrite confirmation for all directions; pre-remove dst to prevent
  adb push nesting source inside existing same-name directory
- F5/F6 ADB→host: show missing destination dialog; FS→ADB: remove duplicate
- F8 on "..": ignore; cycle-copy: sibling-temp workaround
- Both panels refresh after any op

Engine: StatPath (single roundtrip), MoveRemote cp+rm fallback,
live-instance registry replaces vtable-offset trick

Docs: README and .hlf updated for Shift+F5/F6, Ctrl+R, cross-mount


akruphi:
Please, if possible, add "`..`" to devices list to quick return to local computer directory.
Now when I run ADB plugin I have only "`<Not found>`" and to return to host directory I need use **Alt+F12**, but in NetRock I always has "`..`" to quick return to local computer directory.

akruphi:
Also **F9**->Options->Plugins configuration->ADB Plugin not show any dialogs now. Please remove "ADB Plugin" from configuration menu or show configuration dialog.

stpork:
@akruphi  - both concerns should be fixed in 80b158ac26fc62be13b38aa253b9cffe5ba3ce66

PR #3376, opened 2026-04-30
format: replace wide printf UI formatting with FormatString

exkrexpexfex:
Replacing some of swprintf formatting with FormatString builders to increase safety, e.g. the following issues have been fixed 
1. viewer status line rendering failure breaking the test 0005-view-file. can be reproduced by staring far2l with `LC_ALL=C ./far2l --tty` and viewing a file with file path large enough to be truncated with … (U+2026) which is not part of C locale. 
2. far:config display issue on some platforms
image

### DIFF
--- a/far2l/src/cfg/ConfigOptEdit.cpp
+++ b/far2l/src/cfg/ConfigOptEdit.cpp
@@ -164,47 +164,51 @@ public:
 			fsn.Format(L"%-*ls", len_sections_keys, mi.strName.CPtr());
 		}
 		fssave = (_opt.save == OST_COMMON ? "c" : (_opt.save == OST_PANELS ? "p" : "-"));
+		FormatString out;
 		switch (_opt.type)
 		{
-			case ConfigOpt::T_BOOL:
-				mi.strName.Format(L"%s %ls %lc  bool%lc%ls%lc%s",
-					(*_opt.value.b == _opt.def.b ? " " : "*"),
-					fsn.CPtr(), BoxSymbols[BS_V1], BoxSymbols[BS_V1],
-					fssave.CPtr(), BoxSymbols[BS_V1],
-					(*_opt.value.b ? "true" : "false"));
+			case ConfigOpt::T_BOOL: {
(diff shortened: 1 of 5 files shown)


elfmz:
Note that there is setlocale(LC_ALL, ""); in the main.cpp so in my case LC_ALL=C doesnt change anything or may be i m doing something wrong,
In general im ok OK to use FormatString where need to cells accounting for alignments, but for language files let postpone this change, at least this breaks cases when words must be reordered, like:
image
Suggested change for locale issue:
1 is to not use formatting but use ReplaceStrings so order will still be governed by languages, but locale will not affect strings processing. But better to do this as separate commit and ideally - address all language constants, and instead if %ls etc use some neutral patterns, like {1}, {2} etc
2 Or may be better resolve issue with sprintf locale (still unclear why setlocale didnt work for you), as using FormatString everywhere instead of sprintf and co is a bit utopia approach, also printf formatting in many cases more readable.

exkrexpexfex:
Language files changes reverted for later.

I have experimented around and indeed the interface issues are not revealed on Linux when LC_ALL is set to C. setlocale(LC_ALL, "") doesn't redefine it, but takes it again from the environment. Probably macOS standard library is less permissive than glibc here: vswprintf returns -1 and sets errno to 92 (Illegal byte sequence) from within FARStringFmtV when an UTF-8 char is encountered.

For sure, I don't suggest replacing all sprintf calls, only those directly catching real paths or user input. Luckily we don't have many: the two reverted, macro descriptions in macro browser, unlikely plugin names in plugin menu and maybe some paths in far:about.
C-locale is not a big deal, I don't think any user will face such issue anywhere, I surprisingly found it when trying the testing tool. Bigger issue is the coding errors in format strings - the second problem in PR description, which apparently can not be found at compile time, as compilers don't provide printf warnings for wide strings nor support of `__attribute__ ( (format( printf ` or similar, I didn't manage to find any.

elfmz:
ah, its macos, i checked only on Linux

PR #3378, opened 2026-05-01
tree panel: fix passive tree sync and delete refresh

exkrexpexfex:
Fix crash on missing tree updates while directories creation or deletion from filelist panel
image

- keep panel directory on the nearest surviving parent after tree deletes
- refresh passive tree panels on secondary updates and clear stale entries
- fix expanding from the nearest known ancestor

### DIFF
--- a/far2l/src/mix/panelmix.cpp
+++ b/far2l/src/mix/panelmix.cpp
@@ -114,14 +114,24 @@ void ShellUpdatePanels(Panel *SrcPanel, BOOL NeedSetUpADir)
 		if (NeedSetUpADir) {
 			FARString strCurDir;
 			SrcPanel->GetCurDir(strCurDir);
+			while (!strCurDir.IsEmpty()) {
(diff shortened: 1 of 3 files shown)


akruphi:
@elfmz вот этот тоже было бы неплохо принять или есть возражения по коду?

far2l/src/mix/panelmix.cpp
```diff
@@ -114,14 +114,24 @@ void ShellUpdatePanels(Panel *SrcPanel, BOOL NeedSetUpADir)
 		if (NeedSetUpADir) {
 			FARString strCurDir;
 			SrcPanel->GetCurDir(strCurDir);
+			while (!strCurDir.IsEmpty()) {
```
elfmz, line 117: may simplify: while (!strCurDir.IsEmpty() && !apiPathIsDir(strCurDir) && CutToSlash(strCurDir, true)) {}

PR #3383, opened 2026-05-06
SDL font picker now saves font path instead of FC descriptor

exkrexpexfex:
for exact matching on load, to avoid non monospaced version to be chosen 

Also layout independent shortcuts are added


tempergate:
Using path to a font is really bad idea. 
there is no guarantee that it won’t change , and configs would be broken

exkrexpexfex:
The initial issue shall be resolved by fontconfig configuration, the suggested with to absolute path is not feasible for different setups, so I'm closing this. Other improvements from this PR will be sent separately later.

Issue #3089, opened 2025-11-17
Alt-Gr keystrokes does not work (Linux)


jmasweden:
Linux Mint 22.2 X11, FAR2L 2.7 GUI

Any keystroke that needs Alt-Gr like \ $ @ and so on does not work.
Alt-Gr or on keyboard that does not have this key its called Right Alt.

These keys works in Terminal and all other apps, so there must be code in FAR2L that remaps the Alt-Gr key to Alt.

In older versions of FAR2L it worked so this must be a regression.

unxed:
Could you please specify what keyboard layout you're using and whether you have any non-default features enabled in your input settings? Also, could you provide step-by-step instructions on how to reproduce this?

Also, can I ask you to use git bisect to find actual commit which caused the problem.

@elfmz I haven't worked with the right alt key, so I don't really remember how it's handled. I have typographic symbols on it, and everything works fine. The API allows us to distinguish between alt keys, so if we're really having the right alt key remapped to the left one somewhere, and it's interfering with someone's typing, maybe we should get rid of it.

elfmz:
isnt this similar to https://github.com/elfmz/far2l/issues/27 ?

jmasweden:
Normal Swedish keyboard layout. No non-default setting.
I get the same problem on five computers with varying versions of Linux Mint (21/22).
I'm from Sweden so all my computers uses Swedish keyboard settings :)

Similar to #27 but its closed (I did not search for closed/fixed), so must be a regression?


And its getting weird, an almost identical version work (packaged as Appimage):

My installed FAR2L is: 2.7.0-2025-11-17-ca7c79f
Alt-Gr does not work (works as a Alt key)

Then I went to: https://github.com/spvkgn/far2l-portable/releases
And downloaded the appimage version: 2.7.0-2025-11-17-ece4422
Here Alt-Gr works as it should!!!
(but in Appimage privilege escalation does not work so its not as useful)

jmasweden:
Addition, Alt-Gr keys does NOT work in Editor either.
But in terminal (far2l --tty) it works, so the problem is in the WX backend??

jmasweden:
Short update, Latest beta 2.7.0 2020-12-13 (Linux Mint 22, Swedish keyboard)
Now AltGr does NOT work in terminal any more. I tried to find where it happens but I'm to unfamiliar with the code (not good enough programmer...). 
Anyone have a hint where in the code I should look?

unxed:
[...]

Issue #1671, opened 2023-05-06
Add macro support for Right Alt key combos


stpork:
Currently there is no key macro combos defined for RAlt, although the standalone key is defined. That's in contrary to RCtrl, which does have definitions for various key macros. That creates a problem for keyboards with AltGr that use Right Alt to type essential symbols. For example in FI/SE keyboards AltGr (or RAlt) is used to type "@", "\", "|" and many other important symbols. Currently far2l doesn't allow using default system keyboard layout for AltGr and also doesn't make a distinction between Left and Right Alt, which cripples custom definitions in 'key_macros.ini' so that they all work with Left Alt only, ruining the standard keyboard experience for AltGr keyboards. Thus, please either add RAlt combinations to the list of supported keystrokes OR start supporting default keycodes for AltGr-enabled keyboard layouts.

stpork:
Well, the problem I am referring to is much more trivial - I have a FIN/SWE keyboard and cannot use standard AltGr (aka Right Alt) combinations to print standard  symbols ¶@£$∞§|[]≈é and •¡”¥¢‰¶\{}≠¿` (with shift). I am not sure if that is the same as "thirds level layout" but given that I cannot type most useful symbols like @|\[]{} in far2l without AltGr support it make me the whole application pretty useless for anything than just file browsing. At the same time I can use AltGr in standard terminal or in alternatives like MC or Commander One.

unxed:
> cannot use standard AltGr (aka Right Alt) combinations to print standard symbols ¶@£$∞§|[]≈é and •¡”¥¢‰¶\{}≠¿` (with shift)

Looks like a different issue should be opened about it.

stpork:
Just checked with Audacity and it seems to work.

unxed:
For example, how do you type @ sign? `Alt`+`Shift`+`2`?

stpork:
'@' == Right Alt + 2.

Here are standard layouts for FI/SE keyboards

1. No modifiers:
![image](https://github.com/elfmz/far2l/assets/12996347/dfd3e512-f481-424f-8132-0225d83c936e)
2. Right Alt
![image](https://github.com/elfmz/far2l/assets/12996347/0dc48f14-000f-4743-aa50-1a6bc49b2924)
3. Shift + Right Alt
![image](https://github.com/elfmz/far2l/assets/12996347/f172d797-b5d3-4059-9779-ce8b2bffaf44)

unxed:
[...]

Issue #940, opened 2021-02-28
AltGr characters does not work


Pensionarspoolen:
I have configured my system (Mint 20.1 Cinnamon) to have English on screen and swedish keyboard/number settings.
The swedish keyboard has an AltGr key used to typ characters like @$\ and so on. In all other apps and in terminal the AltGr key works BUT in far2l it does NOT work. 
What happens when I try to type AltGr-2 (that should give an @) is that the keybar at the bottom goes blank and no key is typed.
Is it possible to enable/use the AltGr-X keys?

elfmz:
Try uncheck  F9/Interface settings/ Exclusively handle hotkeys that include/Right Alt

elfmz:
Please tell if it help, i'll then make this default.

Pensionarspoolen:
Was not checked, checked save, unchecked save to be sure ;-)
Did not help :-(

unxed:
Confirming.
Linux Mint MATE 20.1 x64, swedish keyboard layout.
AltGr (Right Alt) + 2 does not generate "@" in far2l (but do it in all other apps).

elfmz:
in ubuntu 18 i've added Swedish input source, opened gedited, switched to Swedish input language pressed AltGr +2 and nothing happened.. It works like left Alt. Is there something to do more to make it work?

elfmz:
Aha, found that it should be enabled in gnome-tweaks.

Issue #832, opened 2020-10-30
AltGr + special symbol in cmdline (Ex. Finnish keyboard layout and symbol `@`, to print it you need to press AltGr + 2)


MrKalach:
Far2l version: 2.2.0 1f1b9391-alpha

Description:
On some keyboard layouts, to print special symbols, like `@` and `$`, you need to use `AltGr + some key` combination. Currently, it is impossible to do in cmd line mode. As Far handles `AltGr` as normal `Alt`.

PR #3384, opened 2026-05-06
wxBackend: AltGr (Right Alt) character input on macOS and Linux

stpork:
Adds a "Use Right Alt as AltGr" toggle in System Settings -> Input settings. Default off preserves existing Right-Alt-as-Alt-modifier behavior. When on, Right Alt produces the keyboard layout's AltGr glyphs (e.g. R-Alt+2 = '@' on Finnish/Swedish layouts).

Engages composing mode in KeyTracker on Right Alt (raw kVK_RightOption on Mac; ISO_Level3_Shift / XF86 ALTGR on X11/Wayland). In OnKeyDown, printable keys are routed to OnChar so the layout can translate; in OnChar, the Alt (and on Linux also Ctrl, synthesized by X11 from AltGr) modifier bits are stripped so far2l types the character instead of firing an Alt+key binding. Non-printable keys keep Alt+key behavior.

Setting is plumbed through Opt.UseRightAltAsAltGr and the existing ConfigOpt machinery; the runtime flag lives in WinPort (APIOther.cpp) with default-visibility exports so the dynamically-loaded backend resolves the symbol against the bundle_loader executable.

Bugs:
- Closes / fixes regressions in: #832, #940, #3089, #1671
- Improves: #27
- May address: #2493 

image


far2l/src/cfg/config.hpp
```diff
@@ -521,6 +521,7 @@ struct Options
 	int NoBoxes;
 	int ConsolePaintSharp, ExclusiveCtrlLeft, ExclusiveCtrlRight, ExclusiveAltLeft, ExclusiveAltRight,
 			ExclusiveWinLeft, ExclusiveWinRight;
+	int UseRightAltAsAltGr;
```
akruphi: May be use type  `bool`: `bool UseRightAltAsAltGr;`?

far2l/src/cfg/config.cpp
```diff
@@ -605,6 +607,7 @@ void InputSettings()
 		}
 		ApplyConsoleTweaks();
 		XlatReinit();
+		WinPortSetUseRightAltAsAltGr(Opt.UseRightAltAsAltGr ? TRUE : FALSE);
```
akruphi: if type `Opt.UseRightAltAsAltGr` change to `bool` here we can use variable without conditional operator.

far2l/src/cfg/config.hpp
stpork: i kept **int** to match AddCheckbox(..., BOOL *Value*) and the ConfigOpt table, which registers all the neighbours as int*. Happy to switch to bool if you'd prefer.

far2l/src/cfg/config.cpp
stpork: happy to switch this and previous to bool if you'd prefer — just say the word.

far2l/src/cfg/config.hpp
stpork: Updated in [0d15c73](https://github.com/elfmz/far2l/pull/3384/commits/0d15c731806875e20ab2672b643d3e0752bd8c01)

far2l/src/cfg/config.cpp
stpork: Updated in. [0d15c73](https://github.com/elfmz/far2l/pull/3384/commits/0d15c731806875e20ab2672b643d3e0752bd8c01)

unxed:
I ran this branch on X11 (Xvfb, wxWidgets 3.2.4 / GTK 3) and hit three things.

1. On the Polish layout with the option off (the default value) #3448 comes back: AltGr+a enqueues Ą (U+0104), while current master gives ą. The branch drops `_right_alt`, and the bypass of the layout-dependent workaround in OnKeyDown relied on `RightAlt()`.

2. On Linux the option never engages. Setting `_composing` requires `GetKeyCode() == WXK_ALT`, but ISO_Level3_Shift arrives with `code=0`, so on the Polish layout with `UseRightAltAsAltGr=1` the " COMPOSING" line never shows up in the log and the result is the same Ą. On us,ru with the same option COMPOSING does appear in the log, so the setting itself is read correctly — it is the ISO_Level3_Shift branch that is unreachable. Looks like `|| GetKeyCode() == 0` needs to come back into the condition.

3. The block in wx2INPUT_RECORD that adds `RIGHT_ALT_PRESSED | LEFT_CTRL_PRESSED` to VK_MENU itself for raw=ffea/fe03 is left untouched. Because of it, holding right Alt still wipes the key bar titles even with the option off: VK_MENU goes out with 0x10b.

PR #3382, opened 2026-05-05
MTP plugin: New plugin MTP/PTP for USB-connected devices (Android, cameras, iOS Camera Roll)

stpork:
A new plugin to browse USB-connected MTP/PTP devices (Android, cameras, iOS Camera Roll) directly from the file panel.
- Copy / move / rename / delete in both directions; cross-panel copy on the same device skips the host roundtrip.
- Progress + abort, overwrite prompt (Skip/Newer/Rename), English + Russian.
- Vendored static libmtp 1.1.23 — only system dep is libusb-1.0.
- Linux kernel-claimer kick + macOS icdd dislodge handled automatically.

`-DMTP=OFF` to disable. See `mtp/README.md` for full notes.

PR #3398, opened 2026-05-14
ADB plugin: fix plugin menu title after language switch

### LINKED ISSUE #3396: ADB plugin: garbage in the menu after switching interface language
After switching the far2l interface language, the ADB plugin shows garbage in the menu instead of its name. 

```
  ╔═ Plugin commands ══╗
  ║ Advanced compare   ║
  ║ ArcLite            ║
  ║ Calculator         ║
  ║ Case conversion    ║
  ║ Hex editor         ║
  ║ Image Viewer       ║
  ║ Memo               ║
  ║ NetRocks           ║
  ║ OpenWith           ║
  ║ t archive(s)       ║ <====
  ║ Temporary panel    ║
  ╚═ F4 - set hot key ═╝
```

The problem is in these lines:
https://github.com/elfmz/far2l/blob/dd06e95e1c8ffb6ce26db0c5d5196b3c09b0260f/adb/src/FARPlugin.cpp#L47-L49
The localized string is requested only once during static array initialization, not on every call to GetPluginInfoW.

For example, look at how the OpenWith plugin does it:
https://github.com/elfmz/far2l/blob/dd06e95e1c8ffb6ce26db0c5d5196b3c09b0260f/OpenWith/src/OpenWith.cpp#L531-L534

@stpork

stpork:
Fixes elfmz/far2l#3396

`s_menu_strings` was a static array initialized once at first call, so `Lng(MPluginTitle)` was cached and stale after a language switch. Refresh it on every `GetPluginInfoW()` call instead.

### DIFF
--- a/adb/src/FARPlugin.cpp
+++ b/adb/src/FARPlugin.cpp
@@ -44,7 +44,8 @@ SHAREDSYMBOL void WINAPI GetPluginInfoW(PluginInfo *Info)
 	static const wchar_t *s_disk_menu_strings[] = {L"ADB"};
 	Info->DiskMenuStrings = s_disk_menu_strings;
 	Info->DiskMenuStringsNumber = 1;
-	static const wchar_t *s_menu_strings[] = {Lng(MPluginTitle)};
+	static const wchar_t *s_menu_strings[1];
+	s_menu_strings[0] = Lng(MPluginTitle);
 	Info->PluginMenuStrings = s_menu_strings;
 	Info->PluginMenuStringsNumber = 1;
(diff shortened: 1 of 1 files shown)


spnethw:
It seems the same problem is in PR https://github.com/elfmz/far2l/pull/3382

stpork:
> It seems the same problem is in PR #3382

MTP plugin at mtp/src/FARPlugin.cpp:46-47 already has the fixed pattern:

static const wchar_t* s_menu_strings[1];
s_menu_strings[0] = Lng(MPluginTitle);

PR #3399, opened 2026-05-16
feat(netrocks): add SSH private key passphrase prompt

michaellukashov:
SSH connections using encrypted private keys now prompt the user for a passphrase instead of failing immediately with a generic auth failure.

 ### Changes

 New exception type (Erroring.h, Erroring.cpp)
 - ProtocolKeyPassphraseNeededError — thrown when libssh reports SSH_ERROR while reading an encrypted key file, distinguishing it from actual credential
 failures.

 New IPC message (IPC.h)
 - IPC_PI_KEY_PASSPHRASE_NEEDED — communicated from the background broker thread to the main UI thread.

 Broker forwarding (HostRemoteBroker.cpp)
 - Catches ProtocolKeyPassphraseNeededError and sends the new IPC message, keeping the same pattern as existing auth failure and identity mismatch cases.

 Passphrase dialog (InteractiveLogin.cpp, InteractiveLogin.h)
 - InteractivePassphraseDialog — password-edit dialog with the same visual style and retry-numbering as the existing auth retry dialog.
 - InteractivePassphrase() — public API that shows the dialog and updates the in-memory password on success.

 Retry loop (HostRemote.cpp::ReInitialize())
 - On IPC_PI_KEY_PASSPHRASE_NEEDED, prompts the user up to 3 attempts.
 - On successful entry, retries the connection with the provided passphrase.
 - On exhaustion, falls through to OnBroken() with a generic ProtocolError.
 - Also gates _login_mode on !_password.empty() so the cached passphrase can be reused on subsequent reinitializations.

 Localization (eng.lng, rus.lng, bel.lng)
 - 5 new strings: title, retry title, label text, retry label text, and the passphrase prompt.


akruphi:
After https://github.com/elfmz/far2l/pull/3587 Netrocks moved to `plugins` subdirectory

unxed:
О, спасибо! Только тесты падают и конфликтик надо разрешить

PR #3401, opened 2026-05-19
Readme update for far2l install with the homebrew formula

Vladekk:
I've recently added far2l TTY version to homebrew, which is a very convenient method of installing up-to-date CLI software. This PR updates README with instructions how to install far2l via homebrew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,14 +188,19 @@ debuild
 
 </details>
 
-#### Homebrew (TTY only)
+#### Homebrew options
 
-You can install slightly limited 25MB far2l TTY version via [Homebrew Formula](https://formulae.brew.sh/formula/far2l-tty)
-
-First, install homebrew if not yet installed:
+First, install Homebrew if not yet installed:
 ```sh
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
 ```
+
+##### TTY only, latest release compact version
+
+You can install "compact" slightly limited 25MB far2l TTY version via [Homebrew Formula](https://formulae.brew.sh/formula/far2l-tty)
+
+It has no support for plugins in Python and no AWS S3, SMB protocols in NetRocks due to huge dependency size.
+
 For the far2l binary version run
 ```sh
 brew install far2l-tty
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,8 +208,24 @@ docker build . -l far2l
 docker run -it far2l
 ```
 
-See also [Community packages & binaries](#community_bins)
+#### Homebrew (TTY only)
```
akruphi: I suggest move new subsection before "Docker", immediately after "macOS binaries"

Vladekk: @akruphi changed it and modified MacOS part a bit. Please review.

Vladekk: @akruphi please merge if no further comments

akruphi: > @akruphi please merge if no further comments

Only @elfmz can do merge. We are waiting for his decision.

Vladekk: @akruphi I've added formula for the full version, too. Let me know if this is okay, because this formula is hosted on my Github profile. Homebrew project does not like many versions of the same app.

chrisfinazzo: Does the full version use a separate Tap? 🤔

I don't know the official policy at this point, but it seems easier to only have one - as opposed to several forks - and allow people to customize things to their liking by passing flags as needed.

I have other docs edits that work off of these changes, so I'll wait for those to land before pushing them up.

Vladekk: > Does the full version use a separate Tap? 🤔
> 
> I don't know the official policy at this point, but it seems easier to only have one - as opposed to several forks - and allow people to customize things to their liking by passing flags as needed.
> 
> I have other docs edits that work off of these changes, so I'll wait for those to land before pushing them up.

Yes, full version uses a separate tap.

I don't have much experience,  but my understanding is that homebrew does not allow flags in the main repo.

As for options in the custom tap, right now it seems like an overkill. You either use limited version from the main repo, or full version with just three changes. These changes are python plugins, aws and smb in netrocks.

It seems like an additional job without much benefit. I assume most people either are fine with the limited version, or fine with installing much larger version. Amd version that supports either of the mentioned things is always much larger.

Anyway, I am fine with any approach,  but I don't think I will be able to implement the part with custom flags/options. If it will be decided to go this way, just remove section about the full version.

 I have other things I want to add to far2l, and they are my priority.

akruphi: @Vladekk your information partially correlate with info in Section "Building, Contributing, Hacking"->"macOS build" https://github.com/elfmz/far2l#macos-build If possible, please, collect all unofficial brew tap information together (I have not ideas about preferable location of such info inside README).

Vladekk: @akruphi please take a look. There were some compromises. Notably, macOS gui version does not fit well in the homebrew section. So, I moved "download macOS GUI release" to its own section. Also, link from the "HACKING" is not that great, but I don't see how it can be better.

akruphi: Looks good :+1:. But I've never had Mac and don't know the nuances of Homebrew.

All third-part binaries in section "Community packages & binaries" now, but in your version all Homebrew is logical together  :man_shrugging:.

Vladekk: > Looks good 👍. But I've never had Mac and don't know the nuances of Homebrew.

Honestly, me too.  I'm using Homebrew because I'm using immutable Linux distros. Hopefully, someone with Mac will report any issues.

PR #3414, opened 2026-05-27
fix(vt): Ctrl+letter keys not forwarded to child processes under kitty kb protocol

michaellukashov:
# fix(vt): Ctrl+letter keys not forwarded to child processes under kitty keyboard protocol

## Problem

When kitty keyboard protocol is active in the VT shell, Ctrl+letter combinations (A-Z) generate kitty escape sequences (`ESC[4;5u` for Ctrl+D) instead of raw control bytes (`0x04`). Non-far2l child processes — like any terminal application that doesn't implement kitty keyboard protocol — receive nothing for these keypresses. This breaks Ctrl+D, Ctrl+K, Ctrl+U, Ctrl+A, Ctrl+E in editors, shells, and TUI applications running inside far2l's VT panels.

## Root Cause

`TranslateKeyEvent` in `far2l/src/vt/vtshell.cpp` checks `_kitty_kb_flags` first. When set (child negotiated kitty keyboard protocol), `VT_TranslateKeyToKitty` always wins — even for keys that have trivial legacy encodings. The kitty escape sequences are only understood by terminals implementing that protocol, not by arbitrary child processes.

## Fix

In `TranslateKeyEvent`, after kitty translation produces a result for Ctrl+letter keys (ctrl only, no alt/shift, VK between A-Z), fall through to the legacy `VT_TranslateSpecialKey` path which correctly maps Ctrl+D→`\x04`, Ctrl+K→`\x0b`, etc. All other keys (arrows, F-keys, modifiers, multi-modifier combos) continue to use kitty protocol unchanged.

```cpp
// Ctrl+letter falls through to legacy translation
if (!(ctrl && !alt && !shift
    && KeyEvent.wVirtualKeyCode >= 'A'
    && KeyEvent.wVirtualKeyCode <= 'Z')) {
    return as_kitty;
}
```

## Additional changes

- **TTYRawMode enhancement** — explicit termios flag clearing beyond `cfmakeraw` (CREAD|CLOCAL, VMIN=1/VTIME=0, defensive clearing of IUTF8/IXOFF/IUCLC/ECHOCTL). Added `dup()` failure check. Previously, raw mode could fail silently.
- **FAR2L_TTY_DIAG** — runtime diagnostic (env var, no recompile) prints actual terminal flags after raw mode setup
- **Build fix** — `g_tty_profiler` pointer access (`g_tty_profiler.frames` → `g_tty_profiler->frames`) in `TTYBackend.cpp` WriterThread

## Files changed

| File | Lines | Purpose |
|------|-------|---------|
| `far2l/src/vt/vtshell.cpp` | +7/-2 | **Main fix**: kitty kb Ctrl+letter fallthrough |

## Verification

Tested under WindowsTerminal → SSH → far2l running oh-my-pi (a TUI application). Before: Ctrl+D/K/U/A/E produce no effect. After: all Ctrl+letter keys work correctly (confirmed with `FAR2L_TTY_DIAG=1` and byte-level tracing).

### DIFF
--- a/far2l/src/vt/vtshell.cpp
+++ b/far2l/src/vt/vtshell.cpp
@@ -272,9 +272,29 @@ class VTShell : VTOutputReader::IProcessor, VTInputReader::IProcessor, IVTShell
 			if (grantpt(fd_term)==0 && unlockpt(fd_term)==0) {
 				UpdateTerminalSize(fd_term);
 				const char *slavename = ptsname(fd_term);
-				if (slavename && *slavename)
+				if (slavename && *slavename) {
 					_slavename = slavename;
-				else
+					// Put the slave in non-canonical, non-echoing mode so that
+					// individual bytes reach the shell immediately (no ICANON
+					// line buffering). ISIG is deliberately left SET so that the
+					// kernel still generates SIGINT/SIGTSTP/SIGQUIT for control
+					// bytes — clearing ISIG here would disable signal generation
+					// for any child that does not reset termios itself (cat, sh).
+					// Interactive shells (bash) override this on startup anyway;
+					// this only covers the pre-shell window and non-shell children.
+					int fd_slave = open(slavename, O_RDWR | O_NOCTTY);
+					if (fd_slave != -1) {
+						struct termios ts;
+						if (tcgetattr(fd_slave, &ts) == 0) {
+							ts.c_lflag &= ~(ICANON | ECHO);
+							ts.c_iflag &= ~(IXON | ICRNL | IGNCR | INLCR);
+							ts.c_cc[VMIN] = 1;
+							ts.c_cc[VTIME] = 0;
+							tcsetattr(fd_slave, TCSANOW, &ts);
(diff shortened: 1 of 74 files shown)


unxed:
Выглядит разумно

far2l/src/vt/vtshell.cpp
```diff
@@ -272,9 +272,29 @@ class VTShell : VTOutputReader::IProcessor, VTInputReader::IProcessor, IVTShell
 			if (grantpt(fd_term)==0 && unlockpt(fd_term)==0) {
 				UpdateTerminalSize(fd_term);
 				const char *slavename = ptsname(fd_term);
-				if (slavename && *slavename)
+				if (slavename && *slavename) {
 					_slavename = slavename;
-				else
+					// Put the slave in non-canonical, non-echoing mode so that
```
elfmz, line 277: а это зачем? там же ниже по тексту это режим устанавливается со стороны fd_term

PR #3419, opened 2026-06-02
Ctrl+N for adding new items to User Menu

### LINKED ISSUE #1945: Emulate Insert Key on Mac
I try to insert an entry in User Menu (F2). Has anyone succeed at finding the Insert key on Mac Os? :)
fn+enter, fn+~, fn+shift+~, fn+i do not work for me

### LINKED ISSUE #3418: No way to add new items to F2 menu on laptop without Ins
For example, on Macbooks, but also on any laptops with compact keyboard.
I suggest to add Ctrl/Cmd+N to F2 menu built-in hotkeys.

exkrexpexfex:
Fix #3418
Fix #1945


tempergate:
Works correctly

<s>But there is still f4 and control+f4 keys for edit items. 
Maybe it should be useful to add some Control+E key to replace f4 in the case where f-keys are hard-reachable or absent </s> (not planing)

image

PR #3435, opened 2026-06-21
Replace aws-sdk-cpp with custom S3 implementation based on neon

versus-13:
This PR removes the dependency on aws-sdk-cpp, replacing it with a custom S3 protocol implementation built on neon, which is already used elsewhere in the project. The remaining two commits add functionality that previously came implicitly from aws-sdk-cpp but wasn't explicit in the code: support for the UsePathStyle option to switch between virtual-hosted-style and path-style URL addressing, and reading credentials from ~/.aws along with detecting the http(s) prefix from the hostname parameter.


unxed:
Cool!

akruphi:
Please, if possible, correct information in https://github.com/elfmz/far2l/blob/master/README.md also.

PR #3437, opened 2026-06-22
Help viewer: improvements and bugfixes

spnethw:
**1. Add anchor support in .hlf files**
The help viewer had no way to link to a specific position within a topic — only to the topic as a whole. Add named anchor support:
*   **_Define an anchor:_** Place `@??anchor_name` on its own line immediately before the target text.
*   **_Standard link:_** `~LinkTitle~@TopicId??anchor_name@` scrolls to the specific anchor in a target topic.
*   **_Intra-topic shorthand:_** `~LinkTitle~@??anchor_name@` allows for cleaner navigation within the current topic by omitting the `TopicId`.

**_Constraints:_** Anchor markers must occupy their own line. Topic identifiers must not contain the `??` substring, as it is reserved as the separator between the topic identifier and the anchor.

**2. Strip markup before full-text search**
Full-text search (F7) matched against raw .hlf markup, producing phantom results when the search string matched markup characters rather than visible text — e.g. link targets (`@...@`) or `CtrlStartPosChar` markers. Introduce `Help::SanitizeHelpString()` to strip all .hlf formatting from content lines before passing them to `SearchString()`, resolving the long-standing BUGBUG comment in `Help::Search()`.

**3. Fix out-of-bounds buffer read and markup in search title**
`Help::AddTitle()` embedded the search query into a .hlf-formatted string without escaping, causing the help viewer to misinterpret special characters as markup — e.g. `john.doe@gmail.com` caused garbage to appear after the `@`. Additionally, `OutString()` and `StringLen()` performed an unconditional `Str++` after the `@`-link-skipping loop, reading past the null terminator when no closing `@` was present; the fix moves `Str++` inside the loop so it only runs on a successful match.

**4. Fix scroll reset on URL link activation**
Activating a URL link (https://...) in the help viewer reset the scroll position, dropping the user to the first link in the current topic. The fix sets `IsNewTopic = FALSE` before reloading the current topic when `JumpTopic()` returns FALSE for a URL, so the cursor and scroll position are preserved.

**5. Strip centering prefix '^' from search result title**
When a topic title was defined with a centering prefix (`$^#Title#`), the search results displayed `^Title` instead of `Title`. Remove the leading `^` from `strEntryName` after stripping surrounding whitespace, before the `#` delimiters are removed.


akruphi:
Было бы неплохо описать расширение синтаксиса помощи в `HACKING.md`. Насколько я понимаю - это 2 новых управляющих элемента:
```
New Help Control statements:

`@??anchor_name` (at the beginning of a line) - Named anchor inside topic

`~text~@TopicId??anchor_name@` - Reference to anchor inside a topic
(mandatory both TopicId and anchor_name)
```

Без пояснения не очень понятно - при ссылке внутри текущей темы помощи обязательно в ссылке нужно полностью указывать и код темы и якорь? Просто вариант `~text~@??anchor_name@` недопустим?

spnethw:
> Просто вариант `~text~@??anchor_name@` недопустим?

Отличная идея! 👍
Добавил поддержу сокращённого синтаксиса для ссылок на якоря внутри того же самого топика.

В HACKING.md добавим позже, когда всё устаканится.

PR #3441, opened 2026-06-24
Crest plugin

rampitec:
Temporarily draws a highlighted 'crest' in editor while modifiers pressed (Alt/Ctrl/Shift) or locked (Caps/Num/Scroll).

Can colorize symbol under cursor.

Can draw a ruler at the top with position numbers.

Based on original source by Jury Mamaev.
Adopted for far2l by Stanislav Mekhanoshin.

Known issues:

- After pressing a locker (num/caps/scroll lock) you need to press just anything else because no keyboard event comes to plugin when these are pressed and there is no realistic way to check if these are on or off via FAR API.


rampitec:
It is on by default so people notice it. I can turn it off if you want.

rampitec:
> It is on by default so people notice it. I can turn it off if you want.

Actually turned it off by default. The main reason: not every terminal correctly reports keypresses. Users of some terminals may not appreciate it. So, default is now off.

elfmz:
This functionality is present in colorer plugin, i suggest instead of adding yet another plugin, make it more easily configurable in colorer if needed. Currently its part of shheme settings ('show-cross' setting)

rampitec:
> This functionality is present in colorer plugin, i suggest instead of adding yet another plugin, make it more easily configurable in colorer if needed. Currently its part of shheme settings ('show-cross' setting)

It always been in colorer though it is not the same. Colorer draws crest all the time if enabled, this one is only while you are holding one of the modifiers. I.e., it was written for Withdows long time ago even though colorer was present and both me and original author were using both.

akruphi:
It really may be usefull use cross not always but only when keys pressed. I suggest accept this plugin.

tempergate:
> This functionality is present in colorer plugin, i suggest instead of adding yet another plugin, make it more easily configurable in colorer if needed. Currently its part of shheme settings ('show-cross' setting)

Better to have, than not to have. And it is sometimes useful to disable colorer to gain more performance, Having cross alongside could be usefull

tempergate:
<s>look like a bug
pressing second modifier, e.g. shift will remain cross lines, and wont restore base background color
image

note. Bug is not reproducing when colorer plugin is enabled</s> 
colorer correction option need to be disabled with colorer plug disabled

rampitec:
> colorer correction option need to be disabled with colorer plug disabled

Right, as discussed offline plugin setting is needed to tell if colorer is enabled or not. Unfortunately it requires crest to use 2 different redraw methods.

I will try to improve things by eliminating an option, but that will need 2 not exactly sound checks: if colorer is loaded and if it is enabled. Until then it is governed by the plugin setting.

rampitec:
> colorer correction option need to be disabled with colorer plug disabled

I did better. Option is not needed anymore. The same redraw code now works both with and without colorer. It does a bit more redraw, but does not need a setting and does not need other plugin's config.

akruphi:
Very usefull plugin, especially ruller.
@elfmz please add it to master.

viklequick:
Now I have a ruler in editor. many thanks!

image

akruphi:
After https://github.com/elfmz/far2l/pull/3587 all plugins moved to `plugins` subdirectory

rampitec:
> After #3587 all plugins moved to `plugins` subdirectory

Done

tempergate:
@elfmz , please, we need that plugin in mainline. It is turned off by default, so it won’t disturb anybody who didn’t want it

rampitec:
Some people use the plugin with crest itself disabled, only to display the ruler. For those there is a change: current position is highlighted in the ruler. For others it is just vertical line will extend to the title. W/o crest itself it looks like that:

image

With crest enabled:

image

rampitec:
> This functionality is present in colorer plugin, i suggest instead of adding yet another plugin, make it more easily configurable in colorer if needed. Currently its part of shheme settings ('show-cross' setting)

Arguably colorer may be changed to handle keyboard modifier events. Although I see people using the plugin for the completely different reason: to use the ruler in the title bar. That is for sure beyond colorer, it is even outside of the editor space.

yulian5:
@rampitec It's very useful plugin and definitely needed to be in plugin set. But there are a couple problems better to solve:

- interaction with line numbers (I'm looking into that)
- interaction with selection

Nice to have features (not required for current PR):

- for such kind of interactive plugin it will be nice to have ability to change settings inside of editor (second plugin entry in F11 menu, for example)
- nice to have always visible ruler
- in that case nice do not cover first visible line: shift all rows down or make option to have ruler at the bottom
- nice to have vertical ruler (on right side), it will serve for relative position, also with permanent option
- probably some options/hot keys to turn rulers of/on
- more configuration options: on one key - vertical highlighting/ruler, on another key - horizontal
- option to have only rulers displayed - they are useful by themselves

Challenging option - draw dotted-line crosshair instead of  highlighting, but that is a separate task in the same class as ImageViewer.

Issue #3445, opened 2026-06-30
PR3338 seems to be breaking HexViewer scrolling


flowswitch:
Scrolling the hex view down with PgDn in UTF-8 mode jumps forward in some irregular steps depending on file contents (example: top line offset 0, PgDn, top line offset 4A5, PgDn, top line offset 949). Scrolling up with PgUp works correctly (always jumping in "screen capacity minus 1 line" steps), scrolling with Up/Down arrows is correct too. Switching the codepage to ANSI or OEM fixes the problem.

akruphi:
@unxed was author of https://github.com/elfmz/far2l/pull/3338

unxed:
Try #3451 pls!

akruphi:
PR #3451 now in master. If problem solved, @flowswitch please close this issue

PR #3338, opened 2026-04-09
UTF8 support in viewer (hexitor-based logic)

unxed:
Fixes #3195

PR #3451, opened 2026-07-05
Fix irregular PgDn scrolling in UTF-8 Hex Viewer

unxed:
When scrolling down (PgDn) in Hex mode, `Viewer::ReadString` is used to advance the file pointer to calculate the new offset. For UTF-8 files, `vread` was being called without the `Raw` parameter (defaulting to `false`). This caused it to read logical UTF-8 characters instead of raw bytes. Since UTF-8 characters can be multi-byte, the file position advanced by a variable number of bytes per line instead of strictly 16 bytes, resulting in irregular page jumps.

Explicitly passing `Raw = true` to `vread` when in Hex mode ensures the file pointer advances by exactly 16 bytes per line.

Touch #3445 (regression introduced in #3338)


flowswitch:
Confirming: this PR fixes the problem. Thanks!

PR #3455, opened 2026-07-07
Improve ctrl+F handling in netrocks

russiandesman:
The old pain -- you press ctrl+F and get paths you cannot use anywhere.

The behavior now:
Ctrl+F inserts resolved absolute path without any protocol prefixes -- use it in commands which run remotely, i.e. like "cat <ctrl+f>"

Ctrl+Alt+F inserts path with protocol prefix, more like url, i.e. sftp://user@host:port/path/file; it _might_ be useful sometimes. Also key added to change the output to rsync/sftp compatible form, i.e. user@host:protocol:path/file

As a bonus for those who may need it there is a Ctrl+S -- to put to buffer ssh command leading you to the selected file; usage ctrl+S; then open another terminal and paste.

+ present syntax to differentiate absolute vs relative paths is kind of "local only". Changed it to /~/folder for relative and to /folder for absolute; this syntax is well-known across unix community.

Hope you'll like this


far2l/src/panels/filelist.cpp
```diff
@@ -3796,16 +3796,20 @@ void FileList::CopyNames(bool FullPathName, bool RealName)
...
+	// Fall back to less specific fields for plugins that don't provide them.
+	if (!as_url && Info.CurPath && Info.CurPath[0]) {
+		strDest = Info.CurPath;
+	} else if (Info.CurURL && Info.CurURL[0]) {
 		strDest = Info.CurURL;
 	} else if (Info.CurDir && Info.CurDir[0]) {
 		strDest = Info.CurDir;
 	} else {
-		//fprintf(stderr, "Both CurDir and CurURL are empty or null\n");
+		//fprintf(stderr, "CurPath, CurURL and CurDir are all empty or null\n");
```
akruphi, line 3753: Why no action here? May be inside `else` we need `strDest.Clear()`?

russiandesman: here the normal (non-NetRocks) guys live, strDest already contain "/home/path/whatever"; Cleaning it would mean ctrl+F doesn't work in normal file panels anymore

akruphi:
See also https://github.com/elfmz/far2l/pull/3492

elfmz:
Have you checked that after this change other non-UI functionality tied to plugins paths working correctly in NetRocks:
- File search with content searching in remote files
- Bookmarks in Alt+F1/F2

russiandesman:
> Have you checked that after this change other non-UI functionality tied to plugins paths working correctly in NetRocks:
> - File search with content searching in remote files
> - Bookmarks in Alt+F1/F2

Nope. I didn't know it exists. Any insights are welcome, especially in form "you go here, press this, expected is that".

elfmz:
Avoid file search: go over sftp to some server, navigate to for example source tree of far2l, do alt+F7 and search, will see screen like this:
image
go to any file and press F3 see it opened corrcetly, then F4 - and edit it and save and check again that it was really saved

elfmz:
Regarding bookmarks: again navigate on right panel to some sftp etc directory, press Alt+F2, Press Ins, select free 'slot' in Bookmarks section of menu, Press Ins again,  then exit, may be close far2l and open again, then open Alt+F2 and select that items that must successfully navigate to exactly visited before location.

russiandesman:
Explored both scenarios, everything looks good, files opened ok and written to the expected locations, also quick jumping works too. 

Shall I rebase now or merge master into my branch? What workflow is agreed?

russiandesman:
rebased for clarity

Issue #3402, opened 2026-05-19
[Feature interest investigation] "Open processed view by default" settings checkbox and .md files processing support in the viewer


Vladekk:
I found that far2l has support for rendering some file types via pressing F5 in the viewer. This is called "processing". This is useful for html, pdf etc.

Two things are missing (IMO):
1. Support for .md files (markdown) in the default view preprocessor _view.sh_
2. Support for opening files that can be "processed" in this mode by default, so we can see them already rendered. It might be pretty useful, because view.sh supports a plenty of formats: images, executable  binaries, PGP keys and others.

So, I want to see if there is interest for two features:
1. Adding .md files processing via 'glow' or similar to the default _view.sh_.
2. Adding checkbox "Open processed view by default" to the settings

I already implemented this locally with the help from the LLM, and the patch looks fine. I want to see maintainers opinion if I should create PR for these features (one or both).

akruphi:
It is good idea. Please taking into account that `glow` may be not installed in user system and any external utility in [far2l/bootstrap/view.sh](https://github.com/elfmz/far2l/blob/master/far2l/bootstrap/view.sh) should be checked to existence.

exkrexpexfex:
Regarding the second feature, the similar result can be achieved via menu Commands - File associations by adding glow as custom command for *.md as follows

<img width="836" height="482" alt="Image" src="https://github.com/user-attachments/assets/ffea34fa-5843-4d50-81c8-1eb4343a1980" />

Maybe adding some configuration of internal viewer and editor behavior per file type including opening processed view by default would be nice

PR #3460, opened 2026-07-13
Transformer plugin

exkrexpexfex:
Adds the Transformer plugin—a configurable Unix-style text transformation dispatcher for far2l. It pipes editor selections,  complete editor buffers, or selected panel files through external commands and applies or previews their output, with transactional file replacement, optional backups, ANSI-colored viewer output, and extensible INI presets.

image
image
image
image
image
touch #3402


tempergate:
works correctly, adding new formats to local ~/.config/far2l/plugins/transformer/formats works, tested with adding nix format

[nix.ini.txt](https://github.com/user-attachments/files/30083462/nix.ini.txt)

exkrexpexfex:
some experiments on own cancellable executor have been performed, but no robust crossplatform result has been achieved, so rolled back to fsf.Execute() with force push

PR #3470, opened 2026-07-17
NetRocks: remember edited file directory

### LINKED ISSUE #3453: [Bug] [NetRocks] File is written into incorrect directory
I have project layout like:
```
- A
-- src_file1
- B
-- src_file2
- C
-- build_script
```
on remote server. I connect to this server via ssh protocol of NetRocks plugin. I open `src_file1` search for some identifier in `src_file2`, and then return to `src_file1`. When I try to change it and save, saved file is created at directory `B` instead of being saved into `A`. Similarly, when I cd into `C`, build the project, and then edit `src_file1` again, file is created at `C` instead of `A`. Am I doing something wrong?

I use FAR2L 2.8.0 and SSH uses custom proxy command to connect via [coder](https://github.com/coder/coder), if this is relevant.

exkrexpexfex:
Fix #3453

### DIFF
--- a/NetRocks/src/PluginImpl.cpp
+++ b/NetRocks/src/PluginImpl.cpp
@@ -25,6 +25,23 @@
 
 static std::unique_ptr<ConnectionsPool> g_conn_pool;
 
+namespace
+{
+	std::string LocalFilePath(const wchar_t *directory, const wchar_t *name)
+	{
+		if (!directory || !*directory || !name || !*name) {
+			return {};
+		}
+
+		std::string path = Wide2MB(directory);
+		if (path.back() != '/' && path.back() != '\\') {
+			path += '/';
+		}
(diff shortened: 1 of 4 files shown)


SeTSeR (approved):
Thanks! Checked both cases from issue, all of them work!

akruphi:
After https://github.com/elfmz/far2l/pull/3587 Netrocks moved to `plugins` subdirectory

exkrexpexfex:
@akruphi rebased on top of the current master

elfmz:
Sorry but i think bug was not i NetRocks but in far2l side, check current master please.
BTW It also fixes same problem with working with multiarc.

exkrexpexfex:
@elfmz  yes, it works with current master, thanks

PR #3478, opened 2026-07-18
New themes: attic MDA classic, Hercules+,  and modern theme "Ogres and Goblins"

viklequick:
New themes that remind about 1980th, the times of monochrome displays

MDA classic

image

Hercules Classic

image

image

And the modern theme "retro-motivated", that uses modern web colors from olive and sepia colors.

Ogres and Goblins

image

Note: I have added one line fix that makes color corrections more accurate, by ignoring borders. Otherwise, it makes the border colors that are slightly different to the expected ones.


far2l/src/console/farcolors.cpp
```diff
@@ -268,6 +268,12 @@ static uint64_t assembleColor(RGB& fg, RGB& bg) {
 	return color | FOREGROUND_TRUECOLOR | BACKGROUND_TRUECOLOR;
 }
 
+static bool endsWith(const std::string& s, const std::string& suffix)
```
elfmz: there is StrEndsBy(

viklequick: fixed

PR #3484, opened 2026-07-22
python plugin update

m32:
- disabled OPEN_FINDLIST type plugins in the plugin manager - due to Far2L language limitations, according to which a plugin written in C can only handle this type in one way
- fixed usqlite plugin


spnethw:
Offtopic, but can you clarify the purpose of the PYPLUGIN_THREADED macro and why it's not defined anywhere?

m32:
Python slows down program startups; it takes a long time to start, and unxed tried to start in a new thread, which didn't last. Python must run in the program's main thread, otherwise some plugins won't work.
Generally, this variable should be removed.

PR #3488, opened 2026-07-26
editor: reduce per-line memory usage

exkrexpexfex:
Observation: 
image

Here we try to reduce editor memory use for large files by moving per-line booleans into Flags, sharing editor settings and colors, and lazily allocating dialog-only settings, color data, and word-wrap metadata only when needed.


exkrexpexfex:
Reported effect from the changes when trying the new version on the same environment with the same file:
image

Issue #3486, opened 2026-07-25
[ImageViewer] image not rendered in kitty terminal


spnethw:
When using the **ImageViewer** plugin in far2l with the TTY backend inside the **kovidgoyal's kitty** terminal emulator, the image is not displayed upon opening. It only becomes visible **after a mouse movement or any keypress**. This affects both the initial file open and navigation via PgUp/PgDn.

The issue is specific to fullscreen mode (the plugin's own window). Quick View (Ctrl+Q) works correctly and displays the image immediately.

Tested on: Arch Linux (XFCE), kitty 0.48.1, far2l TTY|k backend.

Not affected:
- GUI|WX backend
- WezTerm terminal
- Konsole terminal
- Quick View panel mode (all terminals)

<img width="1241" height="719" alt="Image" src="https://github.com/user-attachments/assets/4f8a2aa8-26f4-418d-abaa-5633333d0465" />

@unxed

spnethw:
Commenting out this single line in TTYBackend::DispatchOutput() makes the problem disappear entirely:
https://github.com/elfmz/far2l/blob/bcfe026fe1d0631a59631bca9d48da38ee23f222/WinPort/src/Backend/TTY/TTYBackend.cpp#L661

@unxed

unxed:
Вот эта штука решает проблему:

```
diff --git a/ImageViewer/src/ImageAtFull.cpp b/ImageViewer/src/ImageAtFull.cpp
index 5fbd3957..475b5dd0 100644
--- a/ImageViewer/src/ImageAtFull.cpp
+++ b/ImageViewer/src/ImageAtFull.cpp
@@ -9,6 +9,9 @@ class ImageViewAtFull : public ImageView
 	COORD _drag_prev_pos{}, _drag_pending{};
 	bool _dragging{false};
 
+public:
+	bool _first_draw{true};
+
 protected:
 	virtual void DenoteInfoAndPan(const std::string &info, const std::string &pan)
 	{
@@ -59,13 +62,14 @@ protected:
 		FarDialogItemData dd_info = { ws_info.size(), (wchar_t*)ws_info.c_str() };
 		g_far.SendDlgMessage(_dlg, DM_SETTEXT, info_text_dlgid, (LONG_PTR)&dd_info);
 
+		_first_draw = true; // Trigger DN_ENTERIDLE ForceShow to redraw image after text/cursor updates
+
 		ImageView::DenoteInfoAndPan(info, pan);
 	}
 
 public:
[...]

PR #3490, opened 2026-07-27
[WIP] ImageViewer: fix image not rendering in kitty terminal

unxed:
In terminals like kitty and WezTerm, drawing text (even background spaces) over a kitty protocol image erases the image.

The previous workaround using `DN_ENTERIDLE` and `ForceShow()` had two major flaws:
1. It only triggered on the first draw (`_first_draw`), leaving images erased during PgUp/PgDn navigation.
2. `ForceShow()` called `DenoteState()`, which dirtied the dialog items and forced Far to immediately redraw the background, instantly erasing the image again.

This commit introduces `ForceRender()` to push the image to the terminal without dirtying the dialog state. It also hooks into `DN_DRAWDIALOG` to set a flag (`_bgr_drawn`) whenever the dialog is redrawn, allowing `DN_ENTERIDLE` to safely render the image exactly once on top of the finished background.

Touch #3486

### DIFF
--- a/ImageViewer/src/ImageAtFull.cpp
+++ b/ImageViewer/src/ImageAtFull.cpp
@@ -65,7 +65,7 @@ protected:
 public:
 	bool may_select{false};
 	bool full_size{false};
-	bool _first_draw{true};
+	bool _bgr_drawn{false};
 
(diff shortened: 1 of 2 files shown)


spnethw:
Unfortunately, there is no effect. The rendering problem remains.

PR #3492, opened 2026-07-30
NetRocks/filelist: fix Ctrl+F inserting garbage/duplicated path

spnethw:
Fix `Ctrl+F` in the NetRocks plugin panel inserting garbage into the command line when no site is connected (sites list view), or a stale URL from a previous connection after disconnecting from a site.

Also fix a related **self-aliasing bug** in `FileList::PluginGetURL()` that surfaced once the above was fixed: whenever a plugin panel reports both `CurDir` and `CurURL` as empty, `Ctrl+F` would insert a duplicated name instead of the plain item name, since `Name` and `strDest` end up being the same object at the call site.

This is visible on the **NetRocks** sites list root (e.g. _"MySite/MySite"_ or _"\<Create site connection\>/\<Create site connection\>"_), and also in the **tmppanel** plugin, which always reports an empty `CurDir`: adding _"/usr/bin"_ to a temp panel and pressing `Ctrl+F` on it inserts _"/usr/bin//usr/bin"_ instead of _"/usr/bin"_.

### DIFF
--- a/NetRocks/src/PluginImpl.cpp
+++ b/NetRocks/src/PluginImpl.cpp
@@ -75,7 +75,7 @@ public:
 
 PluginImpl::PluginImpl(const wchar_t *path, bool path_is_standalone_config, int OpMode)
 {
-	_cur_dir[0] = _panel_title[0] = _format[0] = 0;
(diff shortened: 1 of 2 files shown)


akruphi:
See also: https://github.com/elfmz/far2l/pull/3455

PR #3514, opened 2026-08-06
ECTL_INSERTTEXT micro optmisation

exkrexpexfex:
ECTL_INSERTTEXT is slow as it's looping through ProcessKey, and unconditional FastShow on KEY_ENTER slows it down even more by unnecessary rendering calls.
Improvement can be seen when calling transformer plugin on a text with many lines.

Further rework of ECTL_INSERTTEXT is needed


exkrexpexfex:
Fixed in master

PR #3520, opened 2026-08-06
Ifx iconv dependency on Mac: use CMake 3.11's "find_package(Iconv)".

HughHoyland:
Otherwise, find_library can pick up a wrong libiconv out of multiple installed ones.


HughHoyland:
My apologies, please also take #3522

PR #3527, opened 2026-08-08
X11, Wayland and TTY: support of two clipboard buffers according to ICCM standard: mouse selection grabs to primary clipboard buffer, middle click pastes it back

viklequick:
Refinement for too old and attic PR#3157.

How it works:

* every mouse selection at editor area or editor copies text to primary clipboard buffer, like X11 standards requires.
* middle click pastes text from primary clipboard in edit fields, editor or command line, when focused.
* Guarded: if mouse pointer is over file list in panel, middle click acts as before (e.g. simulates ENTER).

Both actions can be disabled in interface settings individually, e.g. selection and paste.

Supports:

* X11
* Wayland
* TTY (if OSC52 enabled).

In addition, supports OSC52 read when terminal emulator enables it; unfortunately it is disabled everywhere due to security considerations and even more, such emulators simply ignores it without notice (like Konsole). Maybe in future this way can be useful, because terminal emulators are being adding new restrictions continuously and there is a chance to improve this way in future.


tempergate:
@elfmz , please note that pr when it would be possible. We have lots of troubles with clipboard support. It is good to have at least two main clipboards to be supported. 

@viklequick , please, update the pr to resolve conflict

PR #3528, opened 2026-08-08
Happy birthday to far2l: easter egg with fire effect at the bottom of wx window, plus icon

viklequick:
The effect is very close to DN screen saver.

image

The image has been changed

image

Currently it is embedded into code as there is easter egg,. not the feature to make backgrounds for the wx window.


tempergate:
1. No way to setting and disable
2. Animation is lagging

viklequick:
image

tempergate:
<s>) still no settings
Compare to settings in xfce4-terminal
</s>

2)  file-panel and terminal is ok, but it is covered by image the dialog, that is not good


3) editor should be free from any background as it is working region. 
Kate editor as a reference. Working region should be a solid-color background, as it is text editor. Nothing should destruct from text editing in text editors  


4) if it is only an Easter-egg, 1 week of displaying is too much, if it is just a joke, it should de shown only the first start during user session for example, after displaying the Easter-egg, next far-sessions should be free from it

elfmz:
Sorry too much code and whats worse - code-embedded images just for easter egg. Although having background image as a normal feature of WX (and SDL?) backend can be interesting for some users.. But not from hardcoded image, but as external file, may be some simple pics bundled...  And as easter egg - can replace it with this for example: https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcThCqXRHGEfqo7bsr2gePHIF03BsRnlTxJwLeaD7B_MJA&s=10 (its Roshal, not me, just in case :))

viklequick:
No problem, maybe next summer it will be more interesting. Anyway the celebration date (Aug 15)  is passed out.

PR #3536, opened 2026-08-10
colorer: adjust idleJobAmount dynamically

elfmz:


elfmz:
justification: [colorer-smoke-nl.tar.gz](https://github.com/user-attachments/files/30896223/colorer-smoke-nl.tar.gz)

PR #3542, opened 2026-08-11
FISH+

### LINKED ISSUE #3164: [NetRocks] Implement server-side copy in FISH/SHELL protocol + PowerShell support
Just as #3163:

> For example, it will allow you to work from WSL with the host system's FS at native speed.

, but less complicated implementation possible and SSH server can be installed on recent Windows quite easily.

Will require reimplementing remote.sh in PowerShell.

Advantages comparing to other protocols supporting server side copy:

Over NFS:
- available only in server Windows editions

Over SMB:
- you do not need to setup shares

unxed:
Fixes #3164
Touch #3489

PS: You can also implement #556 on top of this. I have it working. But that requires digging deeper into NetRocks itself, which I was hesitant to do. The `NetRocks/src/Protocol/FISHPLUS/INTEGRATION.md` file explains how to set this up if you're interested! The server-side script I’m committing already supports this.

AI-generated PR details below.


unxed:
# NetRocks: add FISH+ protocol

## What this is

FISH+ is an evolution of the classic fish protocol. Where fish simulates a file
system by running one shell command per question, FISH+ uploads a helper that
stays resident on the far side and answers framed requests over the same
stream.

I designed and implemented it in [f4](https://github.com/unxed/f4), a project
that took a lot of inspiration from far2l, and this is a backport of it.

It is added **next to** the existing `shell` protocol, not instead of it.
Nothing that works today changes behaviour.

## Why it is worth having alongside fish

| | fish (`shell`) | FISH+ |
|---|---|---|
| Directory listing | one command per entry | one round trip for the whole directory |
| Timestamps | second-guessed from `ls` | exact, sub-second where `find` is available |
| Reads | sequential | ranged, so resume and seeking work |
| Writes | whole file | ranged |
| Binary data | escaped/limited | length-framed, base64 only when the channel forces it |
| Odd filenames | fragile | newlines and control characters survive (base64-escaped path lines) |
| Host variety | one code path | probes `find` / GNU `stat` / BSD `stat` / `ls`, and says which it picked |

That last row is the one I would most like reviewed: the helper decides at
connect time what the host actually has, and names the winner on the first line
of every listing, so BusyBox/OpenWrt and the BSDs are handled without a
separate protocol.

## How it is put together

**Transport is `WayToShell` + `FISHPLUS/ways.ini`** — the same machinery the
`shell` protocol already uses. This was a deliberate choice over libssh:

- no new build dependency, so `fish+` exists in every far2l build;
- every way of reaching a shell comes for free (ssh, sshpass, a serial line,
  anything spellable as a command);
- it is the mechanism your users already know from `shell`.

`WayToShell` allocates a pty, which normally destroys binary data. FISH+
survives it because the helper tames the line discipline with POSIX `stty` and
announces `tty` when it succeeded; when it did not, the client sends payload as
base64 rather than bytes a terminal would rewrite. Both paths are exercised by
the tests below.

**The server side is taken from f4 byte-for-byte.** `helper.sh` is not edited
here at all — not even to strip the commands this backport does not call. f4 is
where the protocol is developed, and any local edit would have to be re-applied
by hand on every refresh. `Helpers/UPSTREAM.md` records the upstream commit and
lists the five places where the shell and the C++ client are coupled;
`Helpers/sync-from-f4.sh` makes an update a copy plus a version check.

`helper.sh` is BSD-3-Clause, which is GPL-compatible; the notice is preserved in
the file.

## What I deliberately did not do

The protocol can do considerably more than this patch exposes — server-side
recursive delete, remote grep, a remote line index that opens a 10 GB log
instantly, delta-based saving that turns a one-byte edit of a 100 MB file into
one byte of traffic, and background jobs for tree sizes and duplicate search.

None of it is here, because none of it has anywhere in `IProtocol` to put the
answer, and I did not want to reshape core NetRocks interfaces in a patch
that introduces a protocol. **So this is the minimal useful subset, plus a
document explaining how to unlock the rest.**

`NetRocks/src/Protocol/FISHPLUS/INTEGRATION.md` covers each capability: what the
remote side already offers, what far2l would need, and roughly how large the
change is. It also covers two honest gaps:

- **`ExecuteCommand` is unimplemented** — it expects a pty over a FIFO, and the
  FISH+ request stream is strictly sequential and cannot carry one. It refuses
  rather than silently running the command somewhere unexpected. The document
  describes both the cheap fix (`jstart exec`, non-interactive) and the proper
  one (a second channel).

It also describes how to add libssh as an alternative transport for people who
want a pty-free binary-clean channel, if you would rather have both.

One gap found while writing this was fixed at the source rather than worked
around: protocol v1 could read a symlink but not create one, so `mklink` was
added upstream in f4 and the helper here re-synced. `SymlinkCreate` works, and
refuses politely against an older helper by checking the `ln` feature in the
banner. That is the intended workflow for anything else the remote side turns
out to be missing - `Helpers/sync-from-f4.sh` makes the re-sync a one-liner.

## The second commit

`NetRocks: server-side copy` is separate **so it can be dropped**. It is the one
part that touches shared code.

Copying a file from a NetRocks panel into another directory on the same server
currently downloads and re-uploads it. Moving already avoids that — `OpXfer`
compares `IHost::Identity` and calls `Rename`. This does the same for copying:
an optional `IProtocol::FileCopy` defaulting to `ProtocolUnsupportedError`
(exactly the pattern `ExecuteCommand` already uses), the IPC/IHost plumbing that
mirrors `Rename`, and one branch in `OpXfer` next to `_on_site_move` that clears
the flag after the first refusal.

No existing protocol changes and no existing behaviour changes. Implemented for
FISH+ via the helper's recursive `cp`. SCP could use `cp` and SFTP its
copy-data extension the same way, which is why the interface seems worth more
than the single protocol using it here.

If you would rather not take it, drop commit 2 — commit 1 stands alone.

## Testing

The protocol logic was driven against the real `helper.sh` running in a local
shell, which is the only test that proves the shell side and the C++ side agree
on the wire format. 96 checks, all passing:

- listing, metadata, symlinks, ranged reads, mutations, ownership, timestamps
- a 4 KB binary blob containing every byte value, newlines and NULs included,
  written and read back identical — and the session still in sync afterwards,
  which is what proves the byte counting agrees on both ends
- the base64 write path, same check
- filenames containing spaces, tabs and newlines
- refusals (missing path, relative path rejected by the helper's guard) leaving
  the session usable
- remote grep, remote line index, server-side `cp`
- **every listing backend** (`find`, GNU `stat`, `ls`) and **every read backend**
  (`ddbytes`, `dd`, `tailc`, `cat`)
- the `ls` dialects that only appear on hosts I do not have, by shadowing `ls`
  on `PATH` to emulate BusyBox (`--full-time`) and BSD (`-lanT`), each run under
  `UTC`, `Asia/Kathmandu` and `America/New_York` — all three dialects decode a
  pinned mtime to the same instant

`/bin/sh` was `dash` throughout, deliberately: `bash` reads its input a byte at
a time and hides the bootstrap race that the two-step helper upload exists to
avoid.

What I could **not** test here is the far2l UI integration itself — the panel,
the options dialog, the broker lifecycle — so that is where review effort is
best spent. In particular I would appreciate a second opinion on the
`ConfigureProtocolSHELL.cpp` change: I generalized the existing dialog to take
the `ways.ini` subpath and title rather than copying the file, which seemed
better than a duplicate that would drift, but it does touch `shell`'s dialog.

PR #3552, opened 2026-08-15
editor: "All" button in search dialog - list all occurrences at once

enedzvetsky:
Adds an **"All"** button to the editor's search dialog, like far3 has: instead of jumping to matches one by one, it scans the whole file and shows every occurrence in a list at the bottom of the screen. Each row is `line │ position │ source line`, with the match itself highlighted; <kbd>Enter</kbd> jumps to the selected occurrence (honouring the "Select found" checkbox).
image
image

The button appears in the search dialog only — the replace dialog is unchanged.

### What's in the diff

- **`stddlg`** — `GetSearchReplaceString()` takes an optional `ShowAllButton` and now returns `SEARCHDLG_CANCEL` / `SEARCHDLG_OK` / `SEARCHDLG_ALL` instead of a bare `TRUE`/`FALSE`. The existing callers keep working: the old `TRUE` is `SEARCHDLG_OK == 1`, and Cancel stays the last dialog element, which `SearchReplaceDlgProc` relies on.
- **`vmenu`** — `MenuItemEx` gains `HiliteStart` / `HiliteLength`, which highlight an arbitrary fragment of an item. Unlike `AmpPos` this covers more than one character and works with `VMENU_SHOWAMPERSAND` set; cell offsets are computed the same way the row is drawn, so it is also correct for items rendered through `HiText()`.
- **`editor`** — `SearchAll()` collects the occurrences, `ShowFoundItems()` presents the list, `SelectFoundPattern()` performs the jump.
- **help and language files** — new entries for all 9 languages.

### Behaviour on the awkward paths

- The scan shows the usual progress message and can be interrupted with <kbd>Esc</kbd>. An interrupted scan returns to the editor rather than showing a partial list, since the list's title ("Occurrences: N, lines: M") would otherwise read as a whole-file result.
- Collection stops at 100000 occurrences and reports "Too many occurrences, refine the search string". Without a cap, `All` on a single common character in a large file would try to materialise millions of menu items.
- Nothing found → the usual "Could not find the string" message.

### Testing

Built and exercised on macOS (clang, `-DUSEWX=no -DUSEUCD=no`), driving the TUI through a pty: the list renders with the highlight on the correct occurrence in each row, <kbd>Enter</kbd> jumps to the right line and column, and the too-many / not-found paths show their messages. **I have not built or run this on Linux**, so a check on the primary platform would be welcome.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


akruphi:
В vmenu новый цвет подсветки такого фрагмента захардкорен или из текущих цветов темы какой-то берется?

enedzvetsky:
> В vmenu новый цвет подсветки такого фрагмента захардкорен или из текущих цветов темы какой-то берется?

Не захардкоен, берётся из текущей палитры. Подсветка рисуется цветом Colors[VMenuColorHilite] (или VMenuColorHSelect для выбранного пункта), это те же слоты, которыми VMenu подсвечивает hotkey. Для обычного VMenu (как в списке результатов поиска) они заполняются из COL_MENUHIGHLIGHT / COL_MENUSELECTEDHIGHLIGHT, т.е. цвет настраивается через Options → Colors → Menu, как и остальные цвета меню.

akruphi:
Возможно имеет смысл кратко описать нововедения vmenu в отдельном подразделе [HACKING.md](https://github.com/elfmz/far2l/blob/master/HACKING.md)

tempergate:
everything works correct

image

Issue #500, opened 2019-03-06
Regular expression search requires to escape /


singalen:
`/` is not a special symbol in regex, but regex search in editor won't find it unless I escape it.

randolf:
Actually, it's following a common Perl convention (although Perl allows the use of other characters, most developers use "/").  It's also used in Far Manager to allow for the use of options (this can be found in the documentation available by pressing F1 then selecting "search" beside the "Regular expressions" heading):

Ignore case:  /regex/i
Whole line:  /regex/s
Multiple lines:  /regex/m
Ignore space (except "\ "):  /regex/x

You can also combine options, as in:

/ FOO bar /ix
   This matches "foobar" and "FOOBAR"

singalen:
Thanks for explanation.
That decision feels hacky, though. It's not consistent with the existence of "Match case" checkbox in GUI.

singalen:
Now that I think of it, I'm not satisfied with the explanation.
`s/regex/replacement/options` is a widespread pattern in shell utilities, but only `regex` is what supposed to go into the search text box; everything else should have an option in GUI.
What's the point of having a GUI and having to type options in an obscure language?

PR #3576, opened 2026-08-22
Editor: support multiline regex search and replace

exkrexpexfex:
touch #500
Upd: proper separate handling of CR and LF, added option to reveal line endings in editor with checkboxes in settings and a shortcut Ctrl+Shift+F5


m32:
for the text in the editor
one
two
one
two
I press Shift+Ctrl+F5, then Ctrl+F7
I change `one$two` to `two\none`
it finds nothing and doesn't work as I expected—`one` and `two` swap places

exkrexpexfex:
@m32 the current `$` treatment implementation seems to be conventional and adherent to the standards: `$` is a zero-width end-of-line assertion, it doesn't consume the line-ending characters. See e.g. [POSIX specification](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/xbd_chap09.html) which explicitly provides `e$f` as a valid regex example that should never match.

The following works for me

-      Find:    one$\ntwo
       Replace: two\none

For mixed line endings more sophisticated approach can be used to capture and preserve the line ending

-      Find:    one$(\r\n|\r|\n)two
       Replace: two$1one

m32:
Sounds reasonable. e$f – I didn't know that :(

PR #3579, opened 2026-08-23
NetRocks/FISHPLUS: Windows peer support via PowerShell

sogonov:
**Draft** — the change compiles and the plumbing wires up end-to-end on my Linux/WSL side, but the full end-to-end matrix against a Windows peer is not yet ticked. Opening now so it is reviewable while I work through the checklist at the bottom.

Follow-up to #3542. Extends FISH+ to peers whose login shell is PowerShell rather than a POSIX one — Windows OpenSSH with `DefaultShell=cmd.exe` or `DefaultShell=powershell.exe`.

## What this PR adds

- **`Helpers/helper.ps1`** — byte-for-byte copy from f4 (`plugins/netfox/fishplus/helper.ps1`, commit `dcf89fe`). Same drop-in policy as `helper.sh`. `sync-from-f4.sh` now loops over both files and prints each helper's `F4PROTO`/`$F4PROTO` so drift is easy to spot. `UPSTREAM.md` gains a second table row and one new coupling point covering the `flavor:pwsh` banner tag and the base64 bootstrap variant.
- **`FishPlusScript::BootstrapLinePwshB64()`** — PowerShell counterpart of `BootstrapLine()`. PowerShell has no `read` builtin that could sip the helper off the wire one line at a time, so the whole helper travels inside the bootstrap itself: base64-encoded, one printable-ASCII line, decoded with `.NET` on the far side and handed to `Invoke-Expression`. Same ready-marker split trick that keeps a terminal echo of the bootstrap from looking like the shell answering.
- **`Session::HandshakeOptions`** — carries `helper_path`, `base64_pwsh_bootstrap`, `tty_transport`. The existing `Handshake(helper_path, tty_transport)` signature is a thin shim so no existing caller changes.
- **`Features::Flavor()`** — accessor for the `flavor:pwsh` banner tag. The POSIX helper does not announce a flavor, its absence being the default. Today no code branches on the flavor; the accessor is there for future callers.
- **`ProtocolFISHPLUS`** — `Flavor` site option (`auto` default, `posix`, `pwsh`). Auto tries the POSIX bootstrap first, then the pwsh bootstrap in the same way on wrong-flavor failures (`pty disrupted`, `unexpected handshake banner`, `handshake refused by remote host`, `never reported being ready`, `unsupported protocol version`); if the way itself does not land in a PowerShell prompt and `[SSH_PWSH]` exists in `ways.ini`, jump to it for a third attempt with the PowerShell helper. Any error that does not look like a flavor mismatch propagates as-is. `SSH_PWSH` under Auto skips the POSIX probe, since PowerShell has no error path that would return control quickly if fed the POSIX bootstrap.
- **`ways.ini` `[SSH_PWSH]`** — reaches an interactive PowerShell over ssh via `powershell -NoLogo -NoProfile -EncodedCommand <base64>`. `-EncodedCommand` is one bare token (UTF-16LE base64) which sidesteps every quoting layer between here and the target PowerShell — our `ExplodeCommandLine` splits on unquoted `&`/`|`/`<`/`>` even inside quotes, ssh joins its `argv[N..]` with spaces, and sshd on Windows re-parses the command through the peer's DefaultShell. The encoded blob decodes to `Write-Output 'far2l is ready for fishing'; powershell -NoLogo -NoProfile` — greeting the LOGIN section matches, then a nested interactive PowerShell that inherits stdin and consumes the FISH+ bootstrap. Works whether `DefaultShell` is `cmd.exe` or `powershell.exe`.
- **UI** — `ConfigureProtocolFISHPLUS` grows a `Helper flavor` combobox (`Auto` / `POSIX (helper.sh)` / `PowerShell (helper.ps1)`), routed through the shared SHELL dialog via a new `include_flavor` parameter that stays off for the SHELL protocol. Strings in `eng.lng`, `rus.lng`, `bel.lng`.
- **Windows-shape path normalization** — `Session::EncodePathLine` folds a Windows-shape path (`C:\Users\foo`, `X:relative`, `\\srv\share\rest`) into POSIX shape on the wire (`/c/Users/foo`, `/x/relative`, `//srv/share/rest`). The helper only understands POSIX shape; a Windows user's muscle memory produces `C:\...` and the site config's Directory field takes that verbatim, which used to `no such file` even though `Test-Path` on `C:\` succeeds by accident. POSIX-shape paths pass through unchanged, so no existing peer sees a difference.
- **`INTEGRATION.md` §10** — documents the three moving parts (bootstrap variant, `HandshakeOptions` plumbing, auto probe) with cross-links to §6 (why `ExecuteCommand` and PTY integration are still open) and §8 (base64 bootstrap shape). Also lists what the PR intentionally does not ship.

## What this PR does NOT ship

- **No PowerShell PTY integration in the panel command line.** f4 has this (`PtyShellIntegration`, OSC 133 D return-to-panels detection) but far2l's `IProtocol` has no `ExecuteCommand` for FISH+ today — see §6.
- **No pwsh-specific `ffind`-as-job.** The helper carries the `ffindjob` capability so a future far2l implementation of remote search can use it without touching the helper; today `Client::Find` is not part of `IProtocol`.
- **No recursive directory size** (see follow-ups below).
- **No archive entry into a `.zip` inside a remote panel.** `NetRocks.far-plug-wide` does not export `GetVirtualFindDataW` for any protocol; adding it is orthogonal to this PR.
- **No `SERIAL_PWSH`.** The `stty raw -echo` incantation in `[SERIAL/LOGIN]` is POSIX-only, and PowerShell over a serial line is exotic enough to wait for a motivating use case.

## Related follow-ups (planned, separate PRs)

1. **`unxed/f4`** — add a `dsize`/`totalsize` wire command to both `helper.sh` and `helper.ps1` so the peer computes recursive folder size in one round trip (`du -sb` on POSIX, `Get-ChildItem -Recurse | Measure-Object Length -Sum` on PowerShell). Announced via `feats:dsize`. Bumps `F4PROTO` or lands as a feature-gated extension. Benefits both the f4 Go client and this NetRocks client.
2. **`elfmz/far2l`** — sync both helpers via `sync-from-f4.sh` and teach `ProtocolFISHPLUS::GetSize` (and, later, `IProtocol` if the interface grows a proper `GetRecursiveSize`) to call `dsize` when `_sess->Feats().Has("dsize")` on directories; otherwise the current stat behavior.

Also worth a look, out of scope for now:

- `ExecuteCommand` + panel PTY integration — port from f4's `PtyShellIntegration` (Windows peer cmd routing, OSC 133 D return-to-panels detection).
- `Client::Find` / `ffind`-as-job — expose remote grep to NetRocks when `IProtocol` grows a search interface.
- `[SSH_PWSH]` UX polish once the peer set is broader: authentication-key options, a `PowerShell over cmd fallback` OPT if `DefaultShell` idiosyncrasies come up.

## What I verified locally

- `cmake ..` + `make NetRocks NetRocks-FISHPLUS`: clean build on Ubuntu 24.04 (Linux/WSL2).
- `NetRocks/src/Protocol/FISHPLUS/Helpers/ways.ini` parses cleanly through the real `KeyFileHelper` (small probe I wrote against `libutils.a`); the `[SSH_PWSH]` section round-trips its `Command` and options.
- Manual connect against a Windows peer (`DefaultShell=cmd.exe`) via `[SSH]` + `Flavor=Auto`: two `pty disrupted` failures on `[SSH]` are recognized as wrong-flavor, transport jumps to `[SSH_PWSH]` with the pwsh helper, banner comes back as `FISHPLUS 1 flavor:pwsh ...`, home resolves to `/c/Users/<user>`, folder navigation renders.

### Connect matrix

Against a Windows peer with `DefaultShell=cmd.exe`:

| # | Way | Flavor | Expected |
|---|---|---|---|
| 1 | `SSH` | `Auto` | 2× `pty disrupted` on SSH, jump to `SSH_PWSH` + pwsh succeeds |
| 2 | `SSH_PWSH` | `Auto` | pwsh directly, no POSIX probe |
| 3 | `SSH_PWSH` | `PowerShell` | Same as above, explicit |
| 4 | `SSH` | `POSIX` | Fails with `pty disrupted`, no fallback (correct) |
| 5 | `SSH_PWSH` | `POSIX` | Fails (POSIX bootstrap in PowerShell has no natural error path) |

Against a Linux peer (regression):

| # | Way | Flavor | Expected |
|---|---|---|---|
| 6 | `SSH` | `Auto` | One handshake, POSIX, `flavor 'posix'` in READY log |
| 7 | `SSH` | `POSIX` | Same, without a probe |

### UI

- FISH+ site options dialog shows the `Helper flavor` combobox
- SHELL site options dialog does NOT show it
- `Flavor` value persists across restarts

### Wire operations against a Windows peer (pwsh helper)

- `enum` — folder navigation, listings render
- `info` / `linfo` — attributes dialog fills in
- `read` — F3 view a ~1 MB text file
- `read` — F3 view a ~10 MB binary file, download to local, byte-compare
- `write` — F5 copy a local file to the peer, byte-compare
- `mkdir` — F7
- `rm` — F8 on a file
- `rmdir` — F8 on an empty folder
- `mv` — F6 rename
- `cp` — F5 within the peer
- `utime` — mtime survives a reload
- Path shape — `C:\Users\<user>` in the Directory field, `/c/Users/<user>` too, both resolve
- UNC — `\\server\share\...` and `//server/share/...` both resolve

## Commits

```
254edfd NetRocks/FISHPLUS/Helpers: bring in helper.ps1 from f4
cc83a5c NetRocks/FISHPLUS: HandshakeOptions and a PowerShell bootstrap variant
e48212a NetRocks/FISHPLUS: auto-probe between the POSIX and PowerShell helpers
020c7b9 NetRocks/FISHPLUS: SSH_PWSH way and a Helper Flavor combobox
b7e3857 NetRocks/FISHPLUS/INTEGRATION.md: document Windows peer support
f5fce6a NetRocks/FISHPLUS: cover 'pty disrupted' and jump ways on auto probe
4b6364a NetRocks/FISHPLUS: SSH_PWSH way skips the POSIX probe under Auto
c87ee62 NetRocks/FISHPLUS: fold Windows-shape paths into POSIX shape on the wire
```

Would be happy to squash into one or two commits before merge — the split is here to make review easier.


sogonov:
## Testing checklist

Tick as things are verified end-to-end. Anything left unchecked when marking the PR ready-for-review is a known gap.

### Connect matrix — Windows peer (`DefaultShell=cmd.exe`)

- [ ] `Way=SSH`, `Flavor=Auto` — two `pty disrupted` on `[SSH]`, transport jumps to `[SSH_PWSH]` with pwsh helper, `[FISH+] READY, home '/c/Users/<user>', flavor 'pwsh'`
- [ ] `Way=SSH_PWSH`, `Flavor=Auto` — pwsh directly, no POSIX probe
- [ ] `Way=SSH_PWSH`, `Flavor=PowerShell` — same as above, explicit
- [ ] `Way=SSH`, `Flavor=POSIX` — fails cleanly with `pty disrupted`, no fallback (this is correct)
- [ ] `Way=SSH_PWSH`, `Flavor=POSIX` — fails (POSIX bootstrap in PowerShell has no natural error path; may hang until 20 s handshake timeout or 1000-line motd cap)

### Connect matrix — Linux peer (regression)

- [ ] `Way=SSH`, `Flavor=Auto` — one handshake, POSIX, `[FISH+] READY, ..., flavor 'posix'`
- [ ] `Way=SSH`, `Flavor=POSIX` — same, skips the probe

### UI

- [ ] FISH+ site options dialog shows the **Helper flavor** combobox (Auto / POSIX (helper.sh) / PowerShell (helper.ps1))
- [ ] SHELL site options dialog does NOT show the combobox
- [ ] `Flavor` selection persists across a NetRocks restart

### Wire operations against a Windows peer (pwsh helper)

- [ ] `enum` — navigate `C:\Users\<user>\Documents` and back, listings render
- [ ] `info` / `linfo` — Ctrl+A on a file fills the attributes dialog
- [ ] `read` — F3 view a ~1 MB text file, contents look right
- [ ] `read` — F3 view a ~10 MB binary file, F5 it to local, `cmp` matches source byte-for-byte
- [ ] `write` — F5 a local file to the peer, `cmp` on the round-trip
- [ ] `mkdir` — F7, new folder appears
- [ ] `rm` — F8 on a file
- [ ] `rmdir` — F8 on an empty folder
- [ ] `mv` — F6 rename, name updates
- [ ] `cp` — F5 within the peer (same host, different subfolder)
- [ ] `utime` — mtime survives a reload (F2 or exit + reopen)
- [ ] Path shape — `C:\Users\<user>` in the Directory field resolves; `/c/Users/<user>` also resolves
- [ ] UNC — `\\server\share\...` and `//server/share/...` both resolve (if you have a share to test against)

### Diagnostics if something fails

Run far2l with `FAR2L_STDERR=/tmp/far2l.log ./far2l` and share the `[FISH+]` / `NetRocks:` lines from `/tmp/far2l.log`. That is enough context to reproduce the fault.

unxed:
А это же наш helper.ps1 из f4, который я забыл перенести в тот раз? Велосипед не изобретаем?

sogonov:
да, я его тогда и сел донести досюда, видно по датам коммитов
в минимальном варианте заработало, а дотестить нормально не успел

вроде никто кроме меня helper.ps1 еще не переносил, так что будет полезно даже так влить - windows хоcты через fish+ заведутся на самых базовых операциях точно

sogonov:
вот теперь точно то же helper что в f4

unxed:
Супер, спасибо!

PR #3596, opened 2026-08-27
gitgutter: skip saving editor content for diff in case editor content not modified

elfmz:
..- then  do only diff by git directly 

@exkrexpexfex please check if idea is good

### DIFF
--- a/plugins/gitgutter/src/gitgutter.cpp
+++ b/plugins/gitgutter/src/gitgutter.cpp
@@ -731,7 +731,7 @@ static bool WriteEditorBufferToTemp(std::string &path)
 		if (egs.StringEOL && *egs.StringEOL) {
 			Wide2MB(egs.StringEOL, mb, true);
 		} else if (i + 1 < ei.TotalLines) {
-			mb+= "\n";
+			mb+= '\n';
 		}
 		fwrite(mb.data(), 1, mb.size(), f);
 	}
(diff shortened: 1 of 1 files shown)


exkrexpexfex:
The optimization is a good idea, but comparing against the file on disk is equivalent to comparing against the editor buffer only for HEAD and custom branch baselines. For index, the editor buffer must be compared with the version staged by git add, whereas git diff --cached compares HEAD with the index. For unstaged, the buffer must be compared with the file on disk, whereas ordinary git diff compares the index with the file on disk. In both cases, the gutter would show a different set of changes. I suggest limiting the optimization to the safe baselines:

  	const bool use_file_diff = (ei.CurState & ECSTATE_MODIFIED) == 0
  			&& (g_settings.baseline == "head" || IsCustomBaseline(g_settings.baseline));

  	if (use_file_diff || !TryRunEditorBufferDiff(st, out, effective_baseline)) {

elfmz:
may be like now? (just use file being edited)

exkrexpexfex:
Nice! The baseline handling is correct now, but the file-selection check needs adjustment. ECSTATE_MODIFIED remains set after a file has been edited and saved, so the optimization will no longer be used after the first edit. Maybe using  ECSTATE_SAVED is better to detect that the current buffer has no unsaved changes:

  	FileWithEditorContent(const EditorInfo &ei, EditorState &st)
  	{
  		const bool use_file = (ei.CurState & ECSTATE_SAVED) != 0 && FileExists(st.file);
  		if (!use_file) {
  			if (!WriteEditorBufferToTemp(st.temp_path)) {
  				return;
  			}
  			path = st.temp_path;
  			is_temp_path = true;
  		} else {
  			path = st.file;
  		}
  	}

External operations such as checkout or reset may still change the file on disk without changing the editor buffer. To handle that safely, we can use the editor buffer once after RepoStateChanged() detects a change. Normal updates can continue reusing the disk file.

PR #3603, opened 2026-08-30
QuickView: show per-file-type summary in directory qv

elfmz:


exkrexpexfex:
Nice! Something with dots.. maybe it shall search from the right? Keeping DOTFILE semantics intact
image

tempergate:
Also, we have size of files counting in bytes, but it is not very useful in modern world. 
It would be better to show in Mb, Gb, Tb, depending on the size

exkrexpexfex:
@elfmz in current version of quick view there's an issue with very long names in th panel header, see
image

PR #3473, opened 2026-07-17
Panel menu bar: new menus (navigate, objects, view)

viklequick:
More available commands in panel menu bar, including help and exit

Now the menu is looks like

image

image

image

image

Just collected available hot keys and have added it into menu, to solve the lack of commands to be used.

All commands have their accelarate keys, so the menu handling is really easy.

Issue #1538, opened 2023-03-10
Wanna press Ctrl+~ to get home


czaerlag:
Would be very nice to have a hotkey to quickly change the current directory to the current user's home.
I guess it should be Ctrl+~, similar to Ctrl+\ that takes me to the file system root.
Looks like a trifle, but I dream of it every day! )

shmuz:
Use Folder shortcuts (accessible from the Commands menu).
You can assign 10 folders of your choice to RCtrl+0 ... RCtrl+9.
This feature is described in details in Far Help (select Commands -> Folder shortcuts then press F1).

akruphi:
I agree that this will be very useful features.
Similar suggestion was in [#1220](https://github.com/elfmz/far2l/issues/1220#issuecomment-1025227168).

elfmz:
is ~ always on ` key in all sane layouts? if so i guess it makes sense to create Ctrl+` hotkey otherwise not sure

akruphi:
> is ~ always on `` key in all sane layouts? if so i guess it makes sense to create Ctrl+` hotkey otherwise not sure

In my practice the \`  and ~ located in 1 key in standard US-layout. In https://en.wikipedia.org/wiki/List_of_QWERTY_keyboard_language_variants I see very different situations for regional layouts, and in some situations ~ may be in other key then \`, but I hope that it is not problem if in help will be clearly write 
Change to the home directory   Ctrl+\`

Also in Alt+F1/F2 panel please to ~ add shortcut key \` or if not possible set 2 shortcuts to locations change to \` may be more useful (not sure it discussion).

czaerlag:
> is ~ always on `` key in all sane layouts?
Well, I guess English layout is common for most of keyboards. On my English/Russian keyboard "~" and "`" are on the leftmost key in the first row. (And, by the way, backslash in English layout changes for forward slash in Russian layout, but the hotkey works in both).
> if so i guess it makes sense to create Ctrl+`
Yes, that's what I meant

czaerlag:
> is ~ always on `` key in all sane layouts?

Well, I guess English layout is common for most of keyboards. On my English/Russian keyboard ~ and `` are on the leftmost key in the first row. (And, by the way, backslash in English layout changes for forward slash in Russian layout, but the hotkey works in both).

> if so i guess it makes sense to create Ctrl+`

Yes, that's what I meant

czaerlag:
Thanks a lot! )
[...]

PR #3615, opened 2026-09-03
keybar: palette button at the left corner with available commands all at once

viklequick:
keybar: palette button at the left corner with available commands all at once

Button:

image

Editor:

image

Viewer:

image

Panels:

image

Technically, key bar is alone place that has commands available all time. 

This is not the horizontal menu, because unfortunately it has very limited set of commands and it are available only when menu is active.


tempergate:
very useful feature, especially on compact keyboards and mobile computers, that could work even without keyboard

yulian5:
Hi @viklequick,

That is a very useful feature, however, having all available commands is too much, as some needed commands are in Options, Settings, etc. In my opinion, it would be better to have Favorite Commands: users can select the most useful commands from all existing ones and have them handy. Otherwise, you need to scroll and filter, and as a result, it won't be much better than accessing them from the regular menu. Also, we need a hotkey for that menu (please don't use Ctrl+` [Control + backtick], I'm reserving it for some new exciting features).

akruphi:
> (please don't use Ctrl+` [Control + backtick], I'm reserving it for some new exciting features).

@yulian5 now **Ctrl+`** hardcored in far2l for quick jump to home directory

yulian5:
@akruphi, thanks. In any case, I’ll have to remap a few keys—I just don’t see any other options; for example, Ctrl+F8 through Ctrl+F11—it’s unlikely they get used often. I myself mostly use Ctlr+F3, Ctrl+F6 and Ctrl+F12, I just don't remember the others. It's easier to press Ctrl+F12 and select which sot order you need when guessing what does it mean in the function key bar.

Retraining and building new muscle memory takes a maximum of two to three weeks, so there is no point in spending years pressing awkward three-key combinations instead of just one or two for frequent actions. Besides, if you use these keys only a couple of times a month, the effort involved is virtually nil. Some key assignments made 35 years ago for DOS no longer make much sense today. And a lot of keys are conflicting with the popular modern use forcing users to keep multiple combinations in memory - almost impossible to remember different 3 keys combinations across different apps.

akruphi:
@yulian5 I'm agree that some DOS/Windows 9x legacy not need now and may be painless eliminated, but in Linux in far2l I very often use **Ctrl+\`** to quick jump to home directory - usually several times in day. **Ctrl+\`** is new shortcut introduced only in far2l 2 years ago.

Other my opinion: we should try stay far2l compatible with windows far3 in user side, especially for habitual element locations and keys shortcuts. Because many user work simultaneous in Linux/macOS and Windows.

yulian5:
@akruphi, I also work in Linux and Windows and in my Far 3 version also has home directory key but a different one.  I developed some stuff in in Far 3 and other stuff in far2l and now trying to match features and make unified interface.

akruphi:
Note: quick jump to home by **Ctrl+`** was introduced in far2l in 2023 after https://github.com/elfmz/far2l/issues/1538

viklequick:
> having all available commands

Technically, key bar has a list of "favorite" commands. In CUA minds, there is a "tool bar". 

Unfortunately, due to historical reasons, these commands are splitted to 6 groups  (main, ctrl+, alt+, and combinations). Palette button just enlists them all in one menu list. It is not the list of "all" commands. 

For example, we have a menu bar with commands that are not enlisted in key bar, and these commands are not available in palette as well. Surely, the complete list of commands is too long to place it into simple list.

viklequick:
> we need a hotkey for that menu

Not sure which one can be used. Something like Ctrl+Shift+P?

yulian5:
@viklequick The idea is good and it aligns with modern UI design of search-driven interface. But I'm against the current implementation. It lacks of execution speed and creates cognitive load for experienced user.

We already have such implementation in far2l/Far 3 for config options and I'm sure the most users go first to F9 menu and if they cannot find or don't know what they're looking for then they go to the flat list to search/scroll all available options. 

How often do you use the configuration list yourself, rather than the F9 menu? That is exactly how often your list will be used.

Adding a third way to find a command creates friction. If a user wants to open the history, they now have to decide: Do I press Alt+F8? Do I press F9 and navigate the menu? Or do I click this new palette button?

For daily workflow, modifier keys will always be faster than a search palette.

- Modifier Keys (Muscle Memory): Pressing Alt+F1 (Change Left Drive) or Shift+F4 (Create New File) takes a fraction of a second.
- Palette Search: Opening a palette, typing l, i, n, k to find the "Create Symlink" command, and hitting Enter requires 5 to 6 keystrokes and forces you to visually confirm the search results.

However, the command palette acts as "training wheels" or a quick-reference guide. It solves two specific bottlenecks where modifier keys fail:

- The "Infrequent Command" Problem: If you only use a command once a month (like "File Associations" ), you likely won't remember the shortcut. Or shortcut might not even exist.
- The "Shortcut Conflict" Problem on Linux like Alt+F9 with Gnome or GDE.

I think it should not be a first choice button in function key bar in front of F1 but as menu entry in F9 Commands with assigned hot key as complimentary useful addition. Ctrl+Shift+P is perfectly fine and aligns with all modern tools.

The proposed location for the new button is ideal for frequently used and highly useful functions, rather than for a list of rarely used commands that are hard to find. And in screenshot there are commands like F1, F2, etc which are already accessible by one mouse click or modifier plus mouse click - so what does that leave us?

tempergate:
> * Modifier Keys (Muscle Memory): Pressing Alt+F1 (Change Left Drive) or Shift+F4 (Create New File) takes a fraction of a second.

It is great when you have a keyboard with functional f-buttons. 
otherway, new command pallet wold not remove the old way to reach commands. But it makes it possible to reach on keyboard without f-keys or on a device with touch screen primary

viklequick:
> most users go first to F9 menu and if they cannot find

.. they can upvote for my PR #3473 and PR #3614 respectively. Menu has a minimal set of commands due to historical reasons.

> If a user wants to open the history, they now have to decide: Do I press Alt+F8? Do I press F9 and navigate the menu? Or do I click this new palette button?

You're joking me, right? Of course, user will find his own way, if we'll take him an option.

Btw, palette is designed for mouse. Without it, you need to hold modifier keys just to find the position you need, before click on it. 

It will be nice to give you an opportunity to  take you a look on the proposed changes. It is my "ahead-of-master" build here:  https://download.opensuse.org/repositories/home:/viklequick/ (or you can find it at "main" branch of my far2l repo if you prefer to build from source).

In my build the palette list is enabled by default; in upstream I am conservative and it is turned off.

tempergate:
I don’t see to have “favorites” usefull. I have lot of hosts, and we already have user menu, no need in favorites. 
pallet menu designed to be the same by default

yulian5:
@tempergate There are a lot of problems: terminal limitations, non-standard or national keyboards, etc. Users may create list for their particular usage: you cannot cover all cases and waste time on scrolling/filtering. Any list with more than 10-12 items has a special purpose and it's not intended for a normal usage. Add it to F9->Commands, it'll be useful here. 

Why do you think all menus are organized in sections, submenus and not just a flat big list?

User F2 menu is not the same as selection of built-in favorite commands.

> It is great when you have a keyboard with functional f-buttons.

Always visible top menu bar combined with function key bar provide access to about 90% commands with 2 mouse clicks.
Having multiple ways to execute the same command causes confusion, creates additional cognitive load, makes interface cluttered. 

Having multiple ways to invoke the same command, rather than properly organizing the existing interface, is definitely not the the right approach.

> no need in favorites. pallet menu designed to be the same by default

No, definitely not the same. Favorite has limited number of most used commands visible at once. Palette command list is a big slow way to find something you need with scrolling, finding, filtering.

@tempergate, please answer the same question I asked @viklequick: How often do you use config option list in favor of F9 menu?

tempergate:
> Always visible top menu bar

not an option on small screens. 

> How often do you use config option list in favor of F9 menu?

it is the only way to use commands on a steamdeck for example. It dosn't have internal keyboard, it has only touxhscreen and a mouse emulator, and its screen is 7 inch, and it is too small to display standard keybar
Probably there are couple of touchscreen-based devices with the same problem


To edit "favorites" on every device would be terribly uncomfortable, especially when you do not have physical keyboard at all,, and it would be easier to giveup and find another way

yulian5:
> Always visible top menu bar
> not an option on small screens.

And is having an endless list with a small visible area a good option?

> To edit "favorites" on every device would be terribly uncomfortable

Agreed, there are an options like .ini files or export/import which solve exactly that problem.

tempergate:
> And is having an endless list with a small visible area a good option?

the menu is scrabble, and it possible to reach every element

tempergate:
> Agreed, there are an options like .ini files or export/import which solve exactly that problem.

exporting and importing ini is possible now, just coping them directly. The other problem, you'll need extra steps to bring that files to a new computer. 
But if you do not have a keyboard, it could be a problem. 

I don't understand the problem. The old elements won't be removed, nothing would change in UX. Also there could be just an option to disable all new interface, and continue to use the old one

yulian5:
> I don't understand the problem. The old elements won't be removed, nothing would change in UX. Also there could be just an option to disable all new interface, and continue to use the old one

The problem is simple: wrong UI design. You're trying to place limited usage function as the main UI element.
To have F9 menu in that place will be better. Or favorite commands. Not the function that will be useless for the most of the users.

viklequick:
@yulian5 I see, looks like your description should be moved to the issue regarding user-defined favorites.

By the way, thank you for the idea; maybe someone will add the additions you proposed, once the main code will be merged to master..

yulian5:
_I asked Claude to analyze the discussion and then asked a question:_ 

In "start" button does it better to have a full commands palette or a user's favorite list of commands?

_Here's Claude answer :_

Favorites — not the full list. Here's the actual argument, sharpened from last time:

**Putting the full palette behind the corner button is functionally the same door twice.** If that button opens the identical content Ctrl+Shift+P opens, you haven't added a second capability, you've added a second *trigger* for one capability — which is precisely the redundancy yulian5 is already objecting to in the thread ("Alt+F8? F9 menu? this new button? Adding a third way to find a command creates friction"). Favorites doesn't have that problem, because its payload is genuinely different — small, curated, personal — so the corner button and Ctrl+Shift+P end up solving two different problems instead of offering two knobs for one.

**Speed only works if the list is short.** The whole value of a corner button over a menu is "no scrolling, no typing, just look and click." That only holds if what's behind it is a handful of items. A dozens-to-hundreds-entry palette in that slot forces the same scroll/filter the menu already requires — at which point, per yulian5's question to viklequick, it's genuinely no faster than F9. Favorites is the only content that preserves the button's actual reason for existing.

That said, the two real objections raised against favorites in the thread are worth designing around rather than dismissing:

- **Cold start** — a new user opens the button and it's empty. Ship a small sensible default set (copy, rename, mkdir, delete, find — whatever your 5–8 most-invoked actions are) so day one isn't blank, same pattern as a browser's default bookmarks bar.
- **tempergate's no-keyboard-editing concern** — this is solvable inside the same touch-first interaction the palette already needs to support: open the button, tap an "edit" affordance, tap items to add/remove. No keyboard required, because it's the identical list+tap pattern the full palette uses anyway — you're not building a second UI paradigm for this, just an edit mode on the one you're already building.

One more thing worth adding while you're at it: put a trailing "All commands…" row at the bottom of the favorites popup that opens the full palette. That gives one-tap escalation for "my favorite isn't favorited yet" without making the corner button's *default* view the overwhelming one — it's a drill-down, not a second independent entry point, so it doesn't reopen the redundancy problem above.

PR #3631, opened 2026-09-10
Better user menu coloring

exkrexpexfex:
Special coloring scheme based on bash syntax highlighting with recognition of far2l placeholders !.! !# is being developed in mainstream. Getting ready for it
image


exkrexpexfex:
and working colorer schemes merged to upstream are added as well

